PR #17, opened 2026-04-30
proposal: add /forge hint command for CI fix context injection

eshulman2:
When Forge is stuck in a CI fix loop due to missing domain knowledge, engineers can provide context via /forge hint <message> on the PR. The hint is injected into the next attempt_ci_fix prompt and the attempt counter is reset. The workflow waits for the next gate failure before acting, so if the agent already fixed it, the hint is unused.


JoshSalomon (commented):
There is one use case I know from OCP - the ci fails because it can't get resources to run the ci - this requires some different approach (wait a lot before next try, wait for specific times, and I am not even sure how to handle the counters. so will it be possible to have hints like (retry daily on 4AM, and retry hourly on weekends - without adding to the failure counter every time the ci fails to allocate VMs?

I also don't like the only one hint limitation - see my comments on the file.

JoshSalomon (changes requested):

eshulman2:
@JoshSalomon for failing gates there is the skipped gates and the CI fix don't fire until all tests ran do you think we should add an extra fail safe for another case? the hint is to help the agent in case you know what the issue is on its next iteration and the reason it is one hint limit per run is to avoid confusing it

JoshSalomon:
> @JoshSalomon for failing gates there is the skipped gates and the CI fix don't fire until all tests ran do you think we should add an extra fail safe for another case? the hint is to help the agent in case you know what the issue is on its next iteration and the reason it is one hint limit per run is to avoid confusing it

I am not sure I understand , lets discuss it in the weekly call.

eshulman2:
Converted to #173 after refinement. The issue preserves the human-guided CI repair goal while replacing full-budget resets with one bounded additional attempt per unique hint, supporting multiple audited hints, and separating scheduled infrastructure retries into future retry-policy work.

PR #19, opened 2026-05-03
proposal: add pre-PR validation gate via skill-defined checks (009)

JoshSalomon:
## Summary

Adds proposal 009: an opt-in pre-PR validation gate that runs project-specific checks (lint, tests, type-check, build) in a container before opening a PR.

## Key Design Points

- **Skill-driven**: Projects define checks in a `validate-pre-pr/SKILL.md` with an embedded `yaml:checks` block. No skill = no validation (opt-in, non-breaking).
- **Single container for all checks**: A generated script runs all checks in one lightweight container invocation — no per-check overhead.
- **Separate agent for fixes**: When checks fail, a dedicated `validation_fix` agent container fixes the code using the skill's fix guidance.
- **Validate→fix→re-validate loop**: Capped by `max_cycles` (default 3).
- **Three-level blocking**: `required` (must pass or workflow blocks), `best-effort` (try to fix, proceed with annotation if exhausted), `advisory` (report only, never fix or block).
- **`local_review` unchanged**: Runs once after validation passes for its normal diff quality review. Not part of the validation loop.

## Files Changed

- `proposals/009-pre-pr-validation-gate.md` — full proposal

## Open Questions

See the proposal's Open Questions section for items to discuss before implementation.


proposals/009-pre-pr-validation-gate.md
```diff
@@ -0,0 +1,445 @@
...
+2. `local_review` reviews the git diff for code quality issues but does not execute any validation commands.
+3. After PR creation, `ci_evaluator` detects failures and `attempt_ci_fix` tries to fix them, but each attempt requires a push + full upstream CI re-run.
+
+None of these catch failures before the PR is opened.
+
+## Proposal
+
+### Overview
+
+Add an opt-in `validate-pre-pr` skill that defines structured check commands inside a YAML block. A new `pre_pr_validation` node — inserted between `implement_task` and `local_review` — reads the skill, generates a bash script from the check definitions, and runs all checks in a **single lightweight container** invocation. If blocking checks fail, the node spawns a **separate agent container** (`validation_fix`) with the failure output and the skill's fix guidance to repair the issues. The flow then loops back to `pre_pr_validation` for re-validation. The loop repeats up to `max_cycles` times before either blocking the workflow or proceeding with annotations, as configured per-check via a single `blocking` field (`required`, `best-effort`, or `advisory`).
```
eshulman2, line 33: If we expect the agent to write a bash script to run the local tests and linting tools, why not just allow users to expose a script? what is the benefit of having it in the form of a skill and not a script used by a skill or called pragmatically with the output shared with the fixing agent? I think scripts might offer an easier approach WDYT?

proposals/009-pre-pr-validation-gate.md
```diff
@@ -0,0 +1,445 @@
...
+
+## Motivation
+
+### Problem Statement
+
+Forge's current flow is: implement → local_review (code quality only) → create PR → wait for GitHub CI → fix failures → push → re-trigger CI. Each CI fix cycle takes 10–30 minutes of upstream CI runner time, creates a new commit on the PR, and triggers notification noise for reviewers watching the PR. For projects with 5–10 CI checks, a single PR can accumulate 3–5 fix pushes before CI goes green.
+
+The root cause: Forge never runs the project's own validation suite before pushing. The container agent runs tests at its discretion during implementation, but there is no orchestrator-level gate that enforces "lint passes, types check, tests pass" before a PR is created.
+
+This is especially painful for projects with expensive or slow upstream CI (OpenShift, Kubernetes) where each CI run consumes significant shared infrastructure and a failing PR blocks merge queues.
```
eshulman2, line 19: I agree this is more painful to projects with more complex gates but I'm not sure this proposal would solve it as this would require a complex CI infrastructure access to allow running those before pushing. I think this is a real issue but I'm not sure it is addressed in this proposal.

eshulman2 (commented):
I think this is the right idea but possibly an overcomplicated solution. The problem I see is in the YAML-to-script generation step: the YAML currently contains bash commands, which the agent then reassembles into a script to run. This inverts the natural separation. you're writing code inside a data format and then generating code from it instead of just creating a deterministic script you know for sure its result can verify locally to let the agent/the system use.

IMHO a cleaner approach would be to keep the YAML purely declarative (blocking level, fix guidance, timeouts) and let the user write the validation script directly. For example something like:

```
scripts/
  lint.sh
  unit-tests.sh
validation.yaml:
  checks:
    - name: lint
      script: scripts/lint.sh
      blocking: required
      fix_guidance: "..."
```

In this case the orchestrator runs each script, captures exit codes and output, and uses the YAML metadata for routing. No code generation step. The fix agent receives the failure output and fix guidance as before. The user can test the scripts locally. The YAML describes results, not commands — once you accept that, the generate-script step is no longer required.

Additionally, the proposal doesn't address how to link a validation script/skill to a git repository. Currently, skills are resolved by Jira project, not by GitHub repo. If one Jira project maps to multiple repos, or if validation commands differ per repo, this design doesn't cover it. Supporting per-repo validation may require linking skills (or scripts) to GitHub repos which could be a meaningful feature in its own right but isn't yet implemented.

proposals/009-pre-pr-validation-gate.md
JoshSalomon: I agree - your idea is good.

> Additionally, the proposal doesn't address how to link a validation script/skill to a git repository. Currently, skills are resolved by Jira project, not by GitHub repo. If one Jira project maps to multiple repos, or if validation commands differ per repo, this design doesn't cover it. Supporting per-repo validation may require linking skills (or scripts) to GitHub repos which could be a meaningful feature in its own right but isn't yet implemented.

Regarding this, can't we have a skill in jira point to a repo ($REPO/.forge/skill>.md)>? - alternatively, can't we have 3-layer skill resolution (repo/.forge, jira-project-setting, default)? 
I think this is a big problem, we need skills per repo for multiple reasons, and jira only resolution is not good enough for real-world scenarios.

proposals/009-pre-pr-validation-gate.md
JoshSalomon: It is partially addressed in this proposal, not completely. I rejected the idea of reading the entire CI pipeline from github and executing it because it is too complex, but I think that this can solve some of the issues. For example, I know projects that do not force running unit tests before the CI - IMHO, this is always bad practice, and the more tests we can run locally, the better and faster results we get. So while this is only a partial solution, I think it is better than nothing.

proposals/009-pre-pr-validation-gate.md
eshulman2: skills repo are accessible in the development stage and the agent is aware of them and using but I wonder if we should expect specific skill for that inside the repo (something I assume some projects will reject) I think it would be the best approach allowing both skill inside the repo and another way for repos who are intimidated by adding something like this to the code repo.

proposals/009-pre-pr-validation-gate.md
eshulman2: I agree, just wanted to mention it doesn't cover it 100%

proposals/009-pre-pr-validation-gate.md
JoshSalomon: I think skills give better control than just scripts (I already agreed to use scripts). So if a project decides not to use repo-level skills, they will have less control over the reaction to script failures. This is OK, we can provide reasonable defaults, but projects that want full control will have to add a skill.
BTW - I don't see any reason why projects that use forrge (in contrast to projects that fear adopting AI-generated code) will not agree to add repo level skill for running tests.

proposals/009-pre-pr-validation-gate.md
JoshSalomon: 100% coverage requires us to compete with github, gitlab and similar ci/cd pipelines. This was not my intention.

proposals/009-pre-pr-validation-gate.md
eshulman2: ok, so just to make sure I got what we are going for. are we going for:
projects create:
- scripts
- yaml for config
default skill would:
- read the yaml
- run the scripts
- address the results as configured

OR

projects create:
- scripts
- yaml for config
forge runs as configured in the yaml, and based on the results and with the results creates a container with skills for addressing the failure meaning the default skill would look something like the gate fix node and custom skills can be used for fixing approaches?

I tend to like option 2 more as the running of the checks is enforced and the agent is just tasked with fixing the issue

eshulman2:
Converted to #174 after refinement. The issue adopts repository-owned deterministic scripts referenced by declarative policy, evaluates the trusted base revision, and makes validation independent per repository. This incorporates the review feedback about avoiding shell commands embedded in YAML and Jira-project-only skill resolution.

PR #21, opened 2026-05-04
proposal: S3-compatible artifact store for .forge/ and workspace cleanup

eshulman2:
## Summary

Adds proposal 011 for two related worker reliability problems:

1. **`.forge/` not shared across workers** — each worker writes task files, CI failure reports, and review plans to a local directory. Subsequent workers on the same ticket start blind, forcing redundant re-fetching and breaking multi-stage artifact passing (e.g. `implement_review` cannot read `review-plan.md` written by an earlier analysis step if the workspace was recreated).

2. **Workspaces not cleaned up reliably** — containers run as root, leaving files the host cannot delete without elevated permissions, causing disk waste and state confusion on restart.

### `ForgeArtifactStore`

A thin S3-compatible client that:
- Downloads the `(ticket, repo)` bucket into `.forge/` at worker start
- Uploads `.forge/` back to the bucket after completion
- Deletes the bucket on PR merge or workflow escalation

Bucket names are derived deterministically from the ticket key and repo name — O(1) lookup, no coordination needed. Multi-repo workflows (one feature touching N repos) produce N independent buckets; workers never share across repos.

Feature is **opt-in** via `FORGE_ARTIFACT_STORE_ENDPOINT`. When empty, `ForgeArtifactStore` is a no-op stub — existing deployments are unaffected.

### `force_delete_workspace()`

Replaces `shutil.rmtree` in `teardown_workspace` with `podman unshare rm -rf` to correctly remove root-owned files left by containers.

## Open Questions

- Should uploads be synchronous (blocking) or fire-and-forget?
- Bucket per-ticket vs per-ticket-per-run (rollback support)?
- Cleanup job: inside worker process or separate service?
- Should `.forge/task.json` be excluded from uploads (it is written fresh each run)?


proposals/011-shared-workspace-for-workers.md
```diff
@@ -0,0 +1,116 @@
...
+The `pull_rebase` call remains correct and necessary even with a shared workspace (to handle external pushes from outside Forge), so the current fix is not wasted work.
+
+## Alternatives Considered
+
+| Alternative | Pros | Cons | Why Not |
+|-------------|------|------|---------|
+| Keep per-worker workspaces + pull_rebase | No infrastructure change | `.forge/` not shared; concurrent workers still race | Current state; acceptable short-term |
+| Object storage for `.forge/` only | Lightweight | Doesn't fix push-rejection; adds S3/GCS dependency | Solves half the problem |
+| Git for `.forge/` state | Auditable, replicable | Extremely complex; merge conflicts on every task | Over-engineered |
+| Single-worker-per-ticket guarantee | Eliminates all races | Limits throughput; hard to enforce across restarts | Architecture constraint, not a solution |
```
JoshSalomon: There is another alternative - keep the .forge/ state in redis. I am not sure it is better than other solutions, I put it here just for completeness

proposals/011-shared-workspace-for-workers.md
```diff
@@ -0,0 +1,116 @@
...
+
+#### Worker Coordination and Locking
+
+Shared state introduces concurrent-write risk. Two workers must not commit and push simultaneously. A file-based or Redis-based lock per ticket controls access to the git operations (commit, push, rebase):
+
+```
+/var/forge/workspaces/AISOS-525/.forge/workspace.lock
+```
+
+Workers acquire the lock before any mutating git operation and release it after push. Read-only operations (container analysis, log fetching) do not require the lock.
```
JoshSalomon: You need to make sure that the lock is released on failures (such as container crash).

eshulman2:
this is the wrong approach

PR #24, opened 2026-05-04
proposal: per-node model tier selection (010)

JoshSalomon:
## Summary

Adds proposal 010: per-node model tier selection with symbolic tiers (`fast`, `standard`, `advanced`) mapped to Haiku/Sonnet/Opus.

## Key Design Points

- **3 hardcoded tiers for PoC**: fast (Haiku), standard (Sonnet), advanced (Opus)
- **Sensible defaults**: PRD generation and implementation use advanced; spec/epic/task generation and reviews use standard; PR descriptions and comment sync use fast
- **Single override config**: `NODE_MODEL_OVERRIDES` env var — JSON dict mapping node names to tiers for field testing
- **Minimal code change**: each workflow node calls `resolve_model_for_node()` before invoking the agent; agent and container infrastructure barely change
- **Future path documented**: configurable tier-to-model mapping, user-defined tiers, per-skill overrides, skill-model compatibility metadata

## Files Changed

- `proposals/010-per-node-model-tier-selection.md` — full proposal


proposals/010-per-node-model-tier-selection.md
```diff
@@ -0,0 +1,326 @@
+# Proposal: Per-Node Model Tier Selection
+
+**Author:** jsalomon
+**Date:** 2026-05-04
+**Status:** Draft
+
+## Summary
+
+Forge currently uses a single LLM model for all orchestrator tasks and optionally a second model for container tasks. This proposal introduces three symbolic model tiers — `fast`, `standard`, `advanced` — with per-node assignment, so that simple tasks (PR descriptions, comment classification) use a cheaper/faster model while complex tasks (PRD generation, code implementation) use a more capable one. For the PoC, tiers are hardcoded to Haiku, Sonnet, and Opus respectively, with a central config for node-to-tier mapping that can be overridden for field testing.
```
eshulman2, line 9: any reason as to why use pre-configured symbolic tiers? won't it make us hard-coded claude-only system? what if a gemini model is the model we want to use?

eshulman2 (commented):
I generally agree with the idea behind the feature but I am not sure I like the tiers model, to me it bounds us to hard with claude and doesn't allow the flexibility  to use other model available with vertex-api. I would suggest that we let the mapping be a free formed map with stage -> model name mapping instead of stage -> tier -> pre-configured model

proposals/010-per-node-model-tier-selection.md
JoshSalomon: I structured it this way for simplicity. I will happily add a configuration for the tiers.
@eshulman2 - a question, should the 3 tiers be pre-defined (for the first version)? or should I make this configurable as well?

eshulman2: my general preference is to have no tiers but direct mapping of models, in case I want to use more than 3 models in the process the tier system would limit me, it would also force me to get read more docs to know what those tiers are. IMHO I would suggest using a direct mapping step -> model or at least allow doing something like step -> model/tier to make sure we are allowing direct model calling.

JoshSalomon: I don't like using direct mapping to the model since this requires more maintenance when for some reason (technical, commercial or other) you can't use models anymore. This is why indirection is good. We can decide to have open list of tiers (not just 3) so this is totally user controllable, although for the first phase, I believe this is overkill. 
But having a tier in the middle makes it much easier to make global changes when you stop paying to some provider for another one, or when you have reasons to stop using some model from HF.

eshulman2: how about a double approach, tiers can be either a tier or a direct model name?

eshulman2:
Converted to #175 after refinement. The replacement removes fixed Claude-shaped tiers, uses exact provider-neutral node/skill model targets with project overrides, pins non-secret policy for in-flight workflows, and explicitly separates connection selection from credential resolution and delivery.

PR #26, opened 2026-05-05
drift-detection

eshulman2:


JoshSalomon (commented):
Generating a revised spec is a tricky part - since the original spec was signed by a human and now it is changed. The drift documentation should be clearly marked as a change from the original contract. This is in line with the Jira comment summarization, but since the spec is a doc that can be inspected outside the Jira context, it should also be marked clearly on the spec.

eshulman2:
@JoshSalomon I agree, I wonder how and where it is best to post those drifts in the real code compared to the spec.

PR #38, opened 2026-05-11
docs: add Forge performance metrics proposal

ekuris-redhat:
## Summary
- Adds proposal 011 defining performance metrics for the Forge SDLC orchestrator
- Covers iteration tracking per workflow stage, cost-benefit analysis, and an observability pipeline roadmap
- Defines KPIs, dashboard panels, and success criteria for measuring Forge's impact on development productivity

## Test plan
- [ ] Review proposal content for accuracy and completeness
- [ ] Validate alignment with existing Forge workflow stages and labels


eshulman2 (commented):
I think this requires re-writing there are several core  issues relative to the existing proposals and to the observability design in forge-observability:


- KPIs lack grounding. Targets like "≤2 iterations per stage" or "60–80% time saved" appear without connection to observed data or the codebase. Where do these come from?
- Several proposed features already exist and are implemented. What's missing isn't re-proposing them — it's specifying the tuning needed to hit the targets.
- No separation between metrics and state. Things like prd_iteration_count could be a Prometheus counter, a LangGraph state field, or a Jira field. The proposal doesn't say.
- Underspecified outputs. The weekly report: where does it go, who receives it, what triggers it? The Grafana dashboard: what data sources, what queries?
- ROI and cost-benefit numbers have no measurement methodology. How will feature_token_cost actually be captured? Via Langfuse? Prometheus? The answer changes the implementation significantly.

PR #42, opened 2026-05-12
proposal: Bug Workflow Redesign

eshulman2:
## Summary

Adds proposal 011 for redesigning the bug workflow with five new stages:

- **Triage** — conditional gate that checks for six required fields before analysis begins; posts a targeted comment and pauses if any are missing
- **Container-based analysis** — `analyze_bug` runs in a standard container with repo access, exploring code to produce a grounded RCA with 1–4 fix options
- **Reflection loop** — `reflect_rca` critic agent validates the RCA, looping up to 3× back to analysis if gaps are found
- **Option selection** — `rca_option_gate` presents fix options; user selects with `>option N` prefix or provides feedback for revision
- **Planning + decomposition** — `plan_bug_fix` generates an implementation plan for the chosen approach; `decompose_plan` creates linked Jira Tasks (one per repo) after approval

Everything from `setup_workspace` onward is unchanged.

## Files

- `proposals/011-bug-workflow-redesign.md` — full proposal
- `proposals/README.md` — index updated


eranco74:
Great proposal — the triage → analysis → reflection → option selection → planning pipeline is a solid improvement over the current single-gate flow.

A few ideas worth considering for the stages that come *after* planning:

**1. Explicit reproduction phase before or during analysis.**
The current design has `analyze_bug` explore code to locate the defect, but doesn't attempt to actually reproduce the bug. Adding a structured reproduction step — even inside the analysis container — would ground the RCA in observed behavior rather than inferred behavior. It also catches cases where the bug is already fixed, environment-specific, or non-reproducible, saving the full pipeline from running on a ghost.

**2. Self-review gate between implementation and PR.**
After `implement_bug_fix` completes but before `create_pr`, a review step that re-reads the RCA, the plan, and the actual code changes could catch cases where the fix addresses a symptom rather than the root cause, or where test coverage is incomplete. This is different from the CI evaluator (which checks if tests pass) — it's a qualitative check on whether the fix and its tests actually prove the bug is resolved. Verdicts like "fix is adequate but tests are incomplete" with specific guidance would reduce review round-trips.

**3. Documentation artifacts alongside the PR.**
The pipeline currently ends at PR creation + CI + human review. Generating structured artifacts — a release notes entry, a changelog addition, and a pre-drafted issue update summarizing the root cause and fix — as part of the workflow would reduce the manual overhead after merge. These could be generated from the RCA + plan + implementation notes that already exist in the pipeline state.

**4. PR feedback as a first-class resumable stage.**
After the PR is submitted and reviewers leave comments, re-entering the workflow to address feedback currently means restarting from the human review gate. A dedicated feedback stage that recovers context from the existing artifacts (RCA, plan, implementation notes), maps reviewer comments to specific changes, and tracks which suggestions were addressed vs. declined with rationale would make multi-round reviews smoother — especially when a different agent session handles the feedback.

None of these change the proposed stages — they'd extend the pipeline after `decompose_plan` / `implement_bug_fix` where the current design inherits the existing nodes unchanged.

eshulman2:
Hi @eranco74 thanks for the detailed review:


> **1. Explicit reproduction phase before or during analysis.** The current design has `analyze_bug` explore code to locate the defect, but doesn't attempt to actually reproduce the bug. Adding a structured reproduction step — even inside the analysis container — would ground the RCA in observed behavior rather than inferred behavior. It also catches cases where the bug is already fixed, environment-specific, or non-reproducible, saving the full pipeline from running on a ghost.

I 100% agree this would be a great addition the problem is that in many cases reproduction requires a full deployment of a product to reproduce the bug. for now I added a unit test verification that is added to the RCA. I would need to explore how to implement reproduction in the future.

> **2. Self-review gate between implementation and PR.** After `implement_bug_fix` completes but before `create_pr`, a review step that re-reads the RCA, the plan, and the actual code changes could catch cases where the fix addresses a symptom rather than the root cause, or where test coverage is incomplete. This is different from the CI evaluator (which checks if tests pass) — it's a qualitative check on whether the fix and its tests actually prove the bug is resolved. Verdicts like "fix is adequate but tests are incomplete" with specific guidance would reduce review round-trips.

Added an enhancement to the proposal to modify the existing reviewer node to provide this in the review as well 

> **3. Documentation artifacts alongside the PR.** The pipeline currently ends at PR creation + CI + human review. Generating structured artifacts — a release notes entry, a changelog addition, and a pre-drafted issue update summarizing the root cause and fix — as part of the workflow would reduce the manual overhead after merge. These could be generated from the RCA + plan + implementation notes that already exist in the pipeline state.

Added to the proposal, now the last node would post a comment in jira with a release note it will also add release note text to the PR message.

> **4. PR feedback as a first-class resumable stage.** After the PR is submitted and reviewers leave comments, re-entering the workflow to address feedback currently means restarting from the human review gate. A dedicated feedback stage that recovers context from the existing artifacts (RCA, plan, implementation notes), maps reviewer comments to specific changes, and tracks which suggestions were addressed vs. declined with rationale would make multi-round reviews smoother — especially when a different agent session handles the feedback.

I need to address the different agent issue in a different scope (this is related to existing issue I need to resolve to which I have another proposal)

> None of these change the proposed stages — they'd extend the pipeline after `decompose_plan` / `implement_bug_fix` where the current design inherits the existing nodes unchanged.

I made some changes to the flow let me know what you thing of the updated version

proposals/bug-workflow-redesign.md
```diff
@@ -0,0 +1,449 @@
...
+#### Stage 2: Analysis
+
+**`analyze_bug`** runs in a standard container (same `ForgeAgent` infra as `implement_bug_fix`, full write permissions inside the container, repo access). Receives the full Jira ticket, `triage_missing_fields`, and any `reflection_critique` from a previous iteration.
+
+The agent explores the codebase — clones repos, checks out branches, reads files, inspects git history — to locate the defect. The resulting RCA must include:
+- Confirmed code location (file, function, line range)
+- Mechanism of failure
+- Trace from trigger to symptom
+- **1–4 distinct fix options**, each with title, description, and trade-offs
+- Embedded code snippets sufficient for the critic to validate without independent exploration
```
adalton: **Suggestion: hypothesis-driven investigation + state space enumeration**

The current description says the agent "explores the codebase to locate the defect," but doesn't prescribe *how* it should investigate. Consider requiring the `analyze-bug.md` prompt to instruct the agent to:

1. **Form explicit hypotheses** — list candidate root causes before investigating, rank by likelihood, then test each against the code.
2. **Document eliminated hypotheses** — record which candidates were ruled out and why. This makes the RCA reasoning chain visible to the human at `rca_option_gate` and gives `reflect_rca` something concrete to validate beyond "do the named files exist?"
3. **Enumerate complete state spaces** — when the bug involves states, phases, or conditions, search the codebase to identify ALL possible values rather than assuming. LLMs tend to handle the cases they've seen rather than systematically enumerating possibilities, and an explicit instruction to search reduces partial fixes.

This doesn't change the node contract — it's prompt guidance for how the agent should use its repo access during exploration.

adalton (commented):
Nice proposal — the triage-through-planning pipeline is a strong improvement, and the enhancements added in response to Eran's feedback (qualitative local review, release notes, post-merge summary) round it out well.

I have a few prompt-level suggestions for the analysis and review stages. None of these change the graph structure or state model — they're all about what the autonomous agents should be instructed to do within the phases that already exist.

proposals/bug-workflow-redesign.md
```diff
@@ -0,0 +1,449 @@
...
+- **Reproducibility assessment**: whether the bug can be demonstrated by a unit or integration test in isolation, and if not, why (e.g., requires a running cluster, environment-specific state, specific infra). If a unit-level test is feasible, the agent includes the full source of a minimal failing test in the RCA output — the test is not committed, but the implementing agent uses it as a specification. If not, it documents the conditions under which the bug manifests so a human can verify independently.
```
adalton, line 115: **Suggestion: add historical analysis and confidence level as required RCA output fields**

Two additional fields that would strengthen the RCA output:

1. **When the bug was introduced** — the agent has repo access, so it can `git blame` and inspect commit history. Knowing the specific commit/PR that caused the regression narrows the fix space significantly and gives the human at `rca_option_gate` confidence that the agent found the right code path, not just a plausible-looking one. (e.g., "Introduced in commit `a1b2c3d`, PR #87, 2026-04-15")

2. **Confidence level** (High/Medium/Low with percentage) — not as an escalation trigger (Forge's gates already handle that), but as a quality signal surfaced at `rca_option_gate`. If the RCA says "Confidence: 60%, only one hypothesis tested," the human knows to push back rather than selecting an option. `reflect_rca` could also use it: a low-confidence RCA that hasn't exhausted its hypotheses is a concrete gap worth looping on.

proposals/bug-workflow-redesign.md
```diff
@@ -0,0 +1,449 @@
...
+
+The existing `local_review` node runs a container that inspects the diff and fixes mechanical breaking issues (linting, compilation errors, test failures). For the bug workflow it is enhanced with a bug-specific prompt (`local-review-bug.md`) that additionally receives the RCA, `selected_fix_approach`, and `plan_content` as context.
+
+The container performs two checks in sequence:
+
+1. **Mechanical check** (existing): run linters, type checkers, and the test suite; report any remaining issues.
+2. **Qualitative check** (new): re-read the RCA and plan, then inspect the actual diff and ask:
+   - Does the change address the confirmed root cause, or only a symptom?
+   - Do the new or modified tests actually prove the bug is fixed (i.e., would they have caught this bug before the fix)?
+   - Does the diff match the scope of the approved plan, or has it drifted significantly?
```
adalton, line 254: **Suggestion: expand the qualitative check with concrete review questions**

The three questions here are a good start — especially "does the change address the root cause or only a symptom?" Consider adding a few more concrete checks to the `local-review-bug.md` prompt:

- **"Could someone break this fix without a test failing?"** — this is the single most actionable test-adequacy question, more specific than "do the tests prove the bug is fixed."
- **Completeness across call sites** — if the fix guards or wraps something in one location, are there similar patterns elsewhere that need the same treatment? A fix that addresses 7 of 8 identical call sites is itself a bug.
- **Backward compatibility / rollback safety** — can this change be reverted cleanly if needed?
- **Security basics** — no leaked secrets in the diff, no injection vectors introduced, error messages don't expose internals.

These are low-cost additions to the prompt that turn the review from a judgment call into a concrete checklist, which should improve consistency across different container runs.

proposals/bug-workflow-redesign.md
```diff
@@ -0,0 +1,449 @@
...
+- Trace from trigger to symptom
+- **1–4 distinct fix options**, each with title, description, and trade-offs
+- Embedded code snippets sufficient for the critic to validate without independent exploration
+- **Reproducibility assessment**: whether the bug can be demonstrated by a unit or integration test in isolation, and if not, why (e.g., requires a running cluster, environment-specific state, specific infra). If a unit-level test is feasible, the agent includes the full source of a minimal failing test in the RCA output — the test is not committed, but the implementing agent uses it as a specification. If not, it documents the conditions under which the bug manifests so a human can verify independently.
+
+**`reflect_rca`** runs in a standard container (same infra) because validating that named files and functions exist at the stated locations requires repo access. Receives the RCA text. Validates:
+- Named files and functions exist at the stated locations
+- Failure mechanism is actually possible given the code
+- Fix options are genuinely distinct
+- No unexplained gaps between trigger and symptom
```
adalton, line 121: **Suggestion: have `reflect_rca` also validate hypothesis coverage and historical grounding**

The current validation criteria check structural correctness (files exist, mechanism is possible, options are distinct, no gaps). If the `analyze-bug.md` prompt is updated to require hypothesis-driven investigation and historical analysis (per my other comments), `reflect_rca` could also validate:

- Were multiple hypotheses considered, or did the agent lock onto the first plausible explanation?
- Was `git blame` / commit history consulted to identify when the bug was introduced?

This would catch a common failure mode where the agent finds *a* plausible root cause but not *the* root cause, because it stopped investigating after the first match.

proposals/bug-workflow-redesign.md
```diff
@@ -0,0 +1,449 @@
...
+You are not starting from scratch. The workspace already contains changes from the previous attempt.
+Your job is to address the reviewer's feedback while preserving work that is still correct.
+
+Before making changes:
+1. Read `git diff main` to understand what the previous attempt changed.
+2. Read the reviewer's feedback carefully — it identifies specific gaps, not a general failure.
+3. Decide what to keep (correct changes, useful test infrastructure, refactoring),
+   what to revert (the parts the reviewer identified as wrong), and what to add or replace.
+
+Do not re-implement everything from scratch unless the reviewer explicitly indicated
```
adalton, line 306: **Suggestion: "test the test" bidirectional validation**

The proposal already includes a failing test in the RCA output when feasible (nice touch). Consider also adding an explicit instruction here for the implementing agent to verify that its regression test(s) **fail without the fix and pass with it**.

This is the difference between "I wrote a test" and "I proved the test catches this bug." It could be as simple as: after writing the test, temporarily revert the fix, confirm the test fails, then re-apply. The `local-review-bug.md` qualitative check could also verify this was done (e.g., the commit log or implementation notes mention bidirectional validation).

This is especially valuable in Forge's autonomous context where there's no human watching the agent work — the bidirectional check is a mechanical proof that the test is meaningful.

eranco74 (approved):
/lgtm

eshulman2:
Thanks @adalton — really appreciated the thorough review. All five suggestions are incorporated in the latest commit.

The hypothesis-driven investigation + state space enumeration framing is exactly the kind of discipline the prompt needs — "explore the codebase to locate the defect" was too open-ended and would've led to agents stopping at the first plausible match. Making the hypothesis log a required RCA output field (and having `reflect_rca` validate it) gives us a concrete hook to catch that failure mode rather than hoping the agent self-corrects.

The bidirectional test validation was a good nudge too. Scoped it to agent-written tests only (not the pre-written RCA test, which is already validated red by definition) to avoid the revert-entanglement problem, and had the qualitative reviewer check for evidence in the commit log rather than trying to reproduce it independently.

adalton (approved):

bogdando:
@eshulman2 

> I 100% agree this would be a great addition the problem is that in many cases reproduction requires a full deployment of a product to reproduce the bug

That is true, however, the workflow should avoid situations when "the actual code changes could catch cases where the fix addresses a symptom rather than the root cause". For that, creating a reproducer should be a stop gate for any fixing process to commence. It could be possible to unblock this step with building and providing the reproducer by someone else, and letting agent to know how to verify the fix against the reproducer as a mandatory and unskillpable step of the bugfix workflow

PR #31, opened 2026-05-07
docs: proposal 011 — revision summary on artifact approval

eshulman2:
## Summary

- Adds `proposals/011-revision-summary-on-approval.md` proposing a structured revision history that mirrors the existing Q&A summary (proposal 001)
- Updates `proposals/README.md` index with the new entry

## What the proposal covers

- `revision_history` state field accumulating feedback + agent-generated change notes across all revision rounds
- `post_revision_summary_if_needed()` utility (modeled on `post_qa_summary_if_needed()`) posted to Jira on artifact approval
- Small `regenerate.md` prompt update to emit a one-sentence change note alongside the revised content
- Integration points in `prd_approval.py` and `spec_approval.py`
- Estimated ~2 days total effort across 3 phases

## Open questions (in the proposal)

- Scope: PRD + spec only, or also plan and task revisions?
- Change note source: agent-generated vs. artifact diff?
- Format: separate Jira comment or merged with the Q&A summary?

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #44, opened 2026-05-13
feat: persist .forge/ artifacts in workflow state for cross-worker continuity

eshulman2:
## Summary
- Adds `forge_artifacts: dict[str, dict[str, str]]` to `PRIntegrationState` (shared by both `FeatureState` and `BugState`) to persist container-written `.forge/` files across workspace recreations, keyed by repo name for multi-repo feature support
- Adds `harvest_forge_artifacts` / `restore_forge_artifacts` utilities in `src/forge/workspace/artifacts.py` — each calling node declares which files it harvests, making the mechanism explicit and portable to future workflows
- Wires harvest into `implement_task` (`handoff.md`), `attempt_ci_fix` (`handoff.md` + `fix-plan.md`), and `implement_review` (`review-plan.md`, `review-objections.md`, `handoff.md`)
- Wires restore into `setup_workspace` and `prepare_workspace` (recreation path only) so containers always start with prior artifacts on disk

## Bug Workflow Coverage
The bug workflow's RCA and planning stages (`analyze_bug`, `reflect_rca`, `plan_bug_fix`) use ephemeral temp directories and parse outputs inline into state — no `.forge/` artifacts need to persist across those runs. The implementation stage reuses the shared nodes (`setup_workspace`, `implement_task`, `implement_review`, `attempt_ci_fix`) that already have harvest/restore wired in. `BugState` inherits `forge_artifacts` from `PRIntegrationState` and initializes it to `{}`. No additional changes needed.

## Updated
- Rebased onto main. The `WorkspaceManager.destroy_workspace` fix previously included here (`podman unshare rm -rf` for root-owned cleanup) was dropped — an equivalent fix already landed on main via #141, and main's version is a superset (adds `OSError` fallback handling).
- `_recreate_workspace_from_fork` on main restructured workspace recreation into an atomic build-beside-then-swap; `restore_forge_artifacts` is now wired in after the swap completes, alongside `write_workspace_identity`.
- All 1978 unit tests pass after rebase.

## Test plan
- [ ] Run a feature workflow end-to-end — verify `handoff.md` is harvested after `implement_task` and restored on workspace recreation
- [ ] Trigger a CI fix cycle — verify `fix-plan.md` + `handoff.md` are harvested and available on retry
- [ ] Trigger an `implement_review` cycle — verify `review-plan.md`, `review-objections.md`, `handoff.md` are persisted
- [ ] Run a bug workflow end-to-end — verify implementation stage works with artifact persistence (shared nodes)

[...]

PR #45, opened 2026-05-14
proposal: Workflow Statistics & Automated Reporting

ekuris-redhat:
## Summary
- Adds proposal for automated workflow statistics and reporting for Forge
- Per-ticket Jira summary posted at workflow completion with machine/human time split, token usage, revision counts, and PR links
- Weekly status report aggregating across all active tickets, delivered via CLI, email (Gmail), and Confluence with historical week-over-week trend charts

## Key design decisions
- Stats stored in LangGraph checkpoint state via `StatsState` mixin (no external dependencies)
- Machine time vs human time tracked separately to identify bottlenecks
- Interim summary on blocked/failed, final summary on completion — stats accumulate across retries
- Both Feature and Bug workflows get identical stats treatment

## Related
- Complements PR #31 (Revision Summary on Artifact Approval) — this proposal tracks revision counts as part of broader workflow stats, not as standalone approval-time summaries

## Test plan
- [ ] Review proposal content for technical accuracy against current codebase
- [ ] Validate alignment with existing workflow nodes and state schema
- [ ] Confirm no overlap/conflict with PR #31


proposals/Jira-workflow-stats-reporting.md
```diff
@@ -0,0 +1,325 @@
...
+
+### Detailed Design
+
+#### 1. State Schema: `StatsState` Mixin
+
+A new `StatsState` TypedDict mixin added to `workflow/base.py`, following the existing pattern of `PRIntegrationState`, `CIIntegrationState`, and `ReviewIntegrationState`:
+
+```python
+class StatsState(TypedDict, total=False):
+    langfuse_trace_id: str | None
```
eshulman2: there is no single trace ID in langfuse as the workflow triggers many calls to AI, is this a list? a link? why is it required when langfuse also groups sessions by tickets?

proposals/Jira-workflow-stats-reporting.md
```diff
@@ -0,0 +1,325 @@
...
+    stage_timestamps: dict[str, dict[str, str]]
+    gate_timestamps: dict[str, dict[str, str]]
+    revision_counts: dict[str, int]
+    token_usage: dict[str, int]
+    stage_token_usage: dict[str, dict[str, int]]
+    review_rounds: int
```
eshulman2: why is review rounds required as a separate field? isn't it captured in revision count?

proposals/Jira-workflow-stats-reporting.md
```diff
@@ -0,0 +1,325 @@
...
+| `revision_counts` | `{"prd": 2, "spec": 1, ...}` | Number of regeneration cycles per stage |
+| `token_usage` | `{"input": 12345, "output": 6789}` | Aggregate token consumption |
+| `stage_token_usage` | `{"prd": {"input": ..., "output": ...}, ...}` | Per-stage token breakdown |
+| `review_rounds` | `int` | Number of human review → implement cycles |
+| `workflow_outcome` | `"completed" \| "blocked" \| "failed"` | Terminal status for reporting |
+
+**Machine time vs human time:** Each stage tracks two separate durations. Machine time measures how long Forge actively worked (e.g., generating a spec took 12 minutes). Human time measures how long the workflow waited at an approval gate or review stage (e.g., spec approval took 1 day 4 hours). Human time will often greatly exceed machine time. The number of human interactions - requested revisions, review rounds, approval cycles - can act as a proxy for Forge output quality. The fewer revision cycles there are, the better Forge is doing at producing quality artifacts humans can approve rapidly. Monitoring both metrics allows teams to understand not only where their time is going, but if Forge output quality is increasing over time.
```
eshulman2: 1. this is problematic, lets say a request waited in the queue for four hours, you can't count it, is this acceptable?
2. I'm not sure how this metric will tell them how forge is over time when it is in a statistical void

proposals/Jira-workflow-stats-reporting.md
```diff
@@ -0,0 +1,325 @@
...
+#### 2. Stats Instrumentation
+
+A utility module `workflow/stats/helpers.py` provides four helper functions that nodes call to record data points:
+
+```python
+def record_stage_start(state: dict, stage: str) -> dict    # machine time start
+def record_stage_end(state: dict, stage: str) -> dict      # machine time end
+def record_gate_start(state: dict, gate: str) -> dict      # human wait start
+def record_gate_end(state: dict, gate: str) -> dict        # human wait end
+def record_tokens(state: dict, stage: str, input_tokens: int, output_tokens: int) -> dict
```
eshulman2, line 77: how are you recording tokens? where and how are you catching this?

proposals/Jira-workflow-stats-reporting.md
```diff
@@ -0,0 +1,325 @@
...
+1. Reads stats fields from the checkpoint state
+2. Formats a Jira wiki markup comment
+3. Posts it to the Feature ticket via `JiraClient.add_comment()`
+
+**Trigger points and summary lifecycle:**
+
+- `escalate_to_blocked` — sets `workflow_outcome = "blocked"` or `"failed"`, calls `post_workflow_summary` with `interim=True`
+- `aggregate_feature_status` — sets `workflow_outcome = "completed"`, calls `post_workflow_summary` with `interim=False` (final summary)
+
+**Interim vs final summaries:** When a workflow is blocked or fails, an interim summary is posted immediately, showing all stats accumulated so far. This gives the team context about what happened and where it broke. If the workflow is later resumed via `forge:retry`, stats continue accumulating — they are never reset on retry. When the workflow eventually completes, a final summary is posted covering the entire lifecycle, including time spent blocked and all retry attempts.
```
eshulman2: I don't think posting it on blocked is the right way to let user know where they are and what failed. forge already adds a comment with the error or failure, I don't see how the stats helps them with that

proposals/Jira-workflow-stats-reporting.md
```diff
@@ -0,0 +1,325 @@
...
+| In Progress | Per-ticket row: key, title, current stage, elapsed time |
+| Blocked | Per-ticket row: key, title, blocked stage, duration, error summary |
+| Bottleneck Analysis | Slowest machine stage avg, longest human wait avg, most revised stage, CI first-pass rate |
+| Token Consumption | Totals and percentage breakdown by stage |
+
+#### 5. Weekly Report Delivery
+
+Three output targets from a single report engine:
+
+**CLI (default):**
```
eshulman2: this is for the entire system report? if there is a cli why do we need to integrated into the code? this is a simple query to redis

proposals/Jira-workflow-stats-reporting.md
```diff
@@ -0,0 +1,325 @@
...
+forge weekly-report --format json --output r.json  # JSON for tooling
+```
+
+**Email:**
+```bash
+forge weekly-report --email                        # configured recipients
+forge weekly-report --email --to "mgr@company.com" # override recipients
+```
+
+- Recipients configured via `FORGE_REPORT_EMAIL_TO` env var or project settings
```
eshulman2: -10 on having it as a static var this means no one can subscribe to it without reaching out to admins

proposals/Jira-workflow-stats-reporting.md
```diff
@@ -0,0 +1,325 @@
...
+- Email delivery via Gmail SMTP (`smtp.gmail.com`) or the Gmail API
+- HTML-formatted email matching the CLI report content
+
+**Confluence:**
```
eshulman2: why do we need this specific integration point? what does it add?

proposals/Jira-workflow-stats-reporting.md
```diff
@@ -0,0 +1,325 @@
...
+  - **Token consumption trend** — total input/output tokens per week
+  - **Revision rate trend** — avg revision count per stage across completed tickets
+  - **Throughput trend** — number of tickets completed, blocked, and in-progress per week
+  - **CI fix rate trend** — percentage of tickets passing CI on first attempt
+  - Charts rendered as Confluence chart macros (bar/line charts using the built-in Chart macro) sourced from a data table on the same page. Each weekly report appends its data row to the table, building the historical view incrementally.
+- New `integrations/confluence/client.py` using the Confluence REST API
+
+Flags can be combined: `forge weekly-report --email --confluence` sends email and updates Confluence in one run.
+
+Scheduling is external — any cron job or CI pipeline can invoke the CLI command on a weekly cadence.
```
eshulman2: where is this cron job is configured?

proposals/Jira-workflow-stats-reporting.md
```diff
@@ -0,0 +1,325 @@
...
+* *Total Tokens:* 31,160 input / 22,540 output
+
+{panel}
+```
+
+Stages that were never reached (e.g., Human Review on a blocked ticket) show "—" for all columns. This makes it immediately clear where the workflow stopped.
+
+For blocked or failed outcomes, the header shows `(x) Blocked` or `(x) Failed` with the `last_error` message included.
+
+#### 4. Weekly Status Report
```
eshulman2, line 147: didn't you want a per project report?

proposals/Jira-workflow-stats-reporting.md
```diff
@@ -0,0 +1,325 @@
...
+
+Scheduling is external — any cron job or CI pipeline can invoke the CLI command on a weekly cadence.
+
+### User Experience
+
+**Per-ticket summary — automatic, no user action required:**
+
+When a workflow completes (or is blocked/fails), the team sees a structured summary comment appear on the Jira ticket. Engineers and managers can review cycle times, identify which stages required revisions, and click through to the Langfuse trace for deep observability.
+
+**Weekly report — on-demand or scheduled:**
```
eshulman2: on demand requires user to have access to the entire system redis I would vote against this can be on demand for admins

proposals/Jira-workflow-stats-reporting.md
```diff
@@ -0,0 +1,325 @@
...
+IN PROGRESS
+  PROJ-120  Feature A         at: Implementation    elapsed: 2h 30m
+  PROJ-122  Feature B         at: Spec Approval     elapsed: 1d 4h
+  PROJ-125  Bug C             at: CI Validation     elapsed: 45m
+
+BLOCKED
+  PROJ-118  Feature D         at: CI Validation     blocked: 2d
+  → CI fixes exhausted after 5 attempts (lint-check, type-check)
+
+BOTTLENECK ANALYSIS
```
eshulman2: what does this info give to us?

proposals/Jira-workflow-stats-reporting.md
```diff
@@ -0,0 +1,325 @@
...
+3. **Phase 3: Per-ticket summary** — Implement `workflow/stats/summary.py`, wire into `aggregate_feature_status` and `escalate_to_blocked`, format Jira comment. (~0.5 day)
+4. **Phase 4: Weekly report engine** — Implement `workflow/stats/weekly_report.py` with checkpoint scanning, aggregation, and report formatting. (~1 day)
+5. **Phase 5: CLI command** — Add `forge weekly-report` subcommand with `--days`, `--output`, `--format` flags. (~0.5 day)
+6. **Phase 6: Email delivery** — Implement `integrations/email/client.py`, add `--email`/`--to` flags, SMTP configuration. (~0.5 day)
+7. **Phase 7: Confluence delivery** — Implement `integrations/confluence/client.py`, add `--confluence`/`--space`/`--parent` flags. (~1 day)
+8. **Phase 8: Tests** — Unit tests for helpers, summary formatting, report aggregation. Integration tests for CLI command. (~1 day)
+
+### Dependencies
+
+- [ ] LLM response objects must expose `input_tokens` / `output_tokens` (already available via Anthropic API responses)
```
eshulman2: is it available in vertex as well??

proposals/Jira-workflow-stats-reporting.md
```diff
@@ -0,0 +1,325 @@
...
+- [ ] Redis checkpointer must support scanning all checkpoints (existing `list_checkpoints` helper)
+- [ ] Confluence REST API credentials for Phase 7 (`FORGE_CONFLUENCE_URL`, `FORGE_CONFLUENCE_TOKEN`)
+- [ ] Gmail SMTP credentials for Phase 6 (`FORGE_GMAIL_USER`, `FORGE_GMAIL_APP_PASSWORD`, `FORGE_REPORT_EMAIL_FROM`)
+
+### Risks
+
+| Risk | Likelihood | Impact | Mitigation |
+|------|------------|--------|------------|
+| Token usage not consistently available from all LLM call paths (direct API vs agent vs container) | Med | Med | Audit all LLM invocation paths in Phase 2; use 0 as fallback for paths that don't expose usage |
```
eshulman2: this means that for container runs you won't have any data on nearly anything

proposals/Jira-workflow-stats-reporting.md
ekuris-redhat: You are right. There is no single trace ID for a workflow. Langfuse already groups traces by session using `session_id=ticket_key`. This field is not needed. I will remove `langfuse_trace_id` from the `StatsState`. Instead I will generate the Langfuse session link directly from the ticket key when formatting the summary comment.

proposals/Jira-workflow-stats-reporting.md
ekuris-redhat: You are right, this is redundant. The `revision_counts["review"]` field already captures this. I will drop `review_rounds` as a separate field and use `revision_counts` consistently for all stages including review.

proposals/Jira-workflow-stats-reporting.md
ekuris-redhat: 1+2. You are right, human time should be removed as a metric. The focus should be on tracking the number of iterations — manual interventions and corrections to Forge.

When Forge produces quality output, fewer corrections are needed, and the development process is more streamlined. The number of iterations is what we should be tracking, not time. Fewer manual interventions indicate that Forge is doing its job well.

I have updated the proposal to focus on iteration counts (revisions per stage, CI fix attempts, review rounds) as the primary quality signal. Time tracking is kept only for stage durations (how long Forge actively worked), not human wait time. The `gate_timestamps` field and all gate instrumentation have been removed.

proposals/Jira-workflow-stats-reporting.md
ekuris-redhat: For calls to the Anthropic API we get the token counts from the response object — specifically `response.usage.input_tokens` and `response.usage.output_tokens`. The `log_llm_call` utility in `utils/logging.py` already accepts these parameters but is not wired up yet. This proposal would make that connection happen.

For calls to Vertex AI, we need to verify that `ChatAnthropicVertex` exposes the usage information in LangChain's `response_metadata` on the returned message objects. Both the direct Anthropic API and Vertex AI paths need to be audited during implementation.

proposals/Jira-workflow-stats-reporting.md
ekuris-redhat: I believe that the presentation of the information up to this stage, particularly due to the blocking status, enables informed decision-making and judgment regarding whether it is advisable to pursue the traditional path or proceed with Forge. It provides context about how much work Forge completed before hitting the block — stage durations, iterations, tokens consumed — which helps the team assess the value of retrying vs. taking a different approach.

proposals/Jira-workflow-stats-reporting.md
ekuris-redhat: Yes, the report should be scoped per project/team. I have added a `--project` flag so each team sees only their tickets. Updated the proposal accordingly.

proposals/Jira-workflow-stats-reporting.md
ekuris-redhat: This is a good point — let's discuss this further. The report does need access to checkpoint state deserialization logic (not just raw Redis keys), but you're right that it may not need to be deeply integrated into the core codebase. I'd like to brainstorm the right boundary here.

proposals/Jira-workflow-stats-reporting.md
ekuris-redhat: Agreed, environment variable is too strict. Recipients should be configurable for each project. This can be done through project settings in Jira or a configuration file that teams manage themselves. I will rework the email delivery to get recipients from project-level configuration instead of a fixed environment variable. Updated the proposal accordingly.

proposals/Jira-workflow-stats-reporting.md
ekuris-redhat: The idea was to give stakeholders a way to look at past weekly reports and see how things are going with some charts. It provides a persistent, browsable history of weekly reports with trend visualizations that email alone doesn't offer.

proposals/Jira-workflow-stats-reporting.md
ekuris-redhat: The proposal intentionally leaves scheduling external — "any cron job or CI pipeline can invoke the CLI command." The idea is that teams configure their own cron/CI trigger since Forge itself doesn't have a built-in scheduler.

proposals/Jira-workflow-stats-reporting.md
ekuris-redhat: That is a good point. Direct Redis access should not be required for regular users. The on-demand CLI should be for admins only. Updated the proposal to mark the CLI as admin-only.

proposals/Jira-workflow-stats-reporting.md
ekuris-redhat: The bottleneck analysis is meant to answer: "Where should we focus improvements?" For example, if spec approval consistently takes 1+ days across tickets, that's a signal to look at the approval process. If CI fix rate is low, that points to prompt/tooling quality issues. But if this isn't actionable enough, I can simplify the weekly report to just the ticket-level data and drop the aggregate analysis section. What level of analysis would be useful?

proposals/Jira-workflow-stats-reporting.md
ekuris-redhat: Yes — token usage metadata is available from Vertex AI as well. The Anthropic API on Vertex returns usage data in the response, and LangChain's `ChatAnthropicVertex` exposes it via `response_metadata` on the returned message objects, same as the direct API path. Both paths need to be audited during implementation to confirm consistent behavior, but the data should be there.

proposals/Jira-workflow-stats-reporting.md
ekuris-redhat: This is a problem because the most expensive part is implementation, and this happens in containers. The container shares Langfuse credentials so we can see what is happening in traces, but no metrics flow back to the orchestrator.

We have a few choices:
1. Query the Langfuse API for information about what happened in the container after it is done
2. Have the container write down what happened in a metrics file (e.g., `.forge/metrics.json`) that the orchestrator can read after execution
3. Accept that we will not know everything and only report what the orchestrator sees directly

I think option 2 is the simplest approach — the container already writes to `.forge/task.json`, so adding a metrics file follows the same pattern. The orchestrator reads it after the container exits, just like it reads the task result. This avoids adding an API dependency on Langfuse at report time.

proposals/Jira-workflow-stats-reporting.md
ekuris-redhat: After thinking about it more — keeping it as a forge subcommand is the right call. The report needs access to the checkpoint state schema, deserialization logic, and Redis connection config that already live in the codebase. Extracting it into a standalone script would duplicate those imports and create a maintenance burden. As a subcommand, it naturally has access to everything it needs and stays in sync when the state schema evolves.

proposals/Jira-workflow-stats-reporting.md
eshulman2: I still fail to see the advantage of this, lets say the team sees 2 iterations on PRD how would that change their decision if the spec generation failed? you get a message with the stage you are in and the error with the option to retry, how does knowing how many iterations you did on a previous stage helps you decide on rather to retry or not?

proposals/Jira-workflow-stats-reporting.md
eshulman2: sounds fair

proposals/Jira-workflow-stats-reporting.md
eshulman2: sounds like this can be solved with a mailing list which would also solve the problem of registering to get the mail, the cli sub command you will add will send the mail to a specific mailing list and people can subscribe to it if they want this way they can also see the history. 2 birds with 1 stone.

proposals/Jira-workflow-stats-reporting.md
eshulman2: if this is an admin only tool how can teams configure their own job triggers? this is contradicting

proposals/Jira-workflow-stats-reporting.md
eshulman2: I'm not against aggregative analysis I'm pro reasoning, how would knowing that a ticket took 2 days to review helps improving the process? how to you measure it if you dropped the time measuring?

proposals/Jira-workflow-stats-reporting.md
ekuris-redhat: From my perspective, if we are at a block/fail status and we know how many iterations have occurred, a cost analysis might lead us to decide that it’s not worth continuing with Forge, and that there is actually room for process improvement. The question is: what is the root cause of the failure/blocker, and what are the costs incurred up to this point?

proposals/Jira-workflow-stats-reporting.md
ekuris-redhat: That's a much simpler approach — send to a mailing list and let people self-subscribe. They get delivery and history in their inbox without us building a Confluence integration. I'll drop the Confluence delivery and replace it with mailing list support. The CLI sends to a configured mailing list address, and anyone who wants the report subscribes to the list. This removes an entire integration (Phase 7 / `integrations/confluence/`).

proposals/Jira-workflow-stats-reporting.md
ekuris-redhat: You're right, that's contradictory. If the CLI requires Redis access then only admins can run it, so the admin would also set up the cron schedule. Teams wouldn't configure their own triggers — they'd request the report cadence from the admin, or the admin sets a default schedule per project. I'll update the proposal to make this consistent: the admin configures the cron job, teams subscribe to the mailing list to receive the output.

proposals/Jira-workflow-stats-reporting.md
ekuris-redhat: This is inconsistent with the decision to drop time tracking. Since we're focused on iterations, the analysis should be: which stages have the highest avg iteration count, what's the CI first-pass rate, and what's the overall avg iterations per completed ticket. Time-based bottleneck metrics should be removed since we dropped `gate_timestamps`. The iteration data directly points to where Forge needs improvement ,if spec has 2.5 avg iterations, that's where to invest in better prompts or context.

proposals/Jira-workflow-stats-reporting.md
eshulman2: we can do option 2 use the interface introduced in https://github.com/forge-sdlc/forge/pull/44 to collect the data from the file. make sure you are collection it on every iteration and sum the data so it is traceable.

proposals/Jira-workflow-stats-reporting.md
ekuris-redhat: We discussed and agreed on the path. I updated the PR according to our conversation.

proposals/Jira-workflow-stats-reporting.md
ekuris-redhat: Updated the PR according to the accepted solution

proposals/Jira-workflow-stats-reporting.md
```diff
@@ -0,0 +1,281 @@
+# Proposal: Workflow Statistics & Automated Reporting
+
+**Author:** Eran Kuris
+**Date:** 2026-05-14
+**Status:** In-Progress
+
+## Summary
+
+When a Forge workflow reaches a terminal state — successful completion, blocked, or unrecoverable failure — the system should automatically post a development statistics summary as a Jira comment on the Feature ticket. The primary quality signal is iteration count: how many manual interventions and corrections were needed at each stage. When a workflow is blocked due to escalation to human, the full stats summary is posted so the team can assess the work Forge completed before hitting the block. If token consumption also exceeds a configurable threshold, a cost alert is included. Users can request stats on demand for any ticket via `/forge stats` (Jira comment) or `forge-watch stats PROJ-123` (CLI). A per-ticket email with aggregate stats is sent to the ticket's reporter and assignee on workflow completion. Additionally, a weekly status report aggregates statistics across all active and completed tickets per project, delivered via admin CLI.
```
danchild: There are some really interesting ideas in here, especially posting a comment to notify us that something is in a "terminal" state in the sense that action needs to be taken. 

Regarding the statistics, I think that a dedicated free form JIRA field would be the best place to display the analysis, using JIRA markup syntax for readability/formatting. Instead of only populating the analysis following a terminal state, we could update the results after each workflow step completes (even driven through redis).  In addition to your metrics you mentioned, we can also include the scorecard concept here. The JIRA comment can explain a subset of the analytics in the case that there was a serious error, explaining what happened. If it's a grave error, we can link these comments to the idea of "alerts" that we will be triggering. The comment could be something like "see alert x in the dashboard to investigate). Then the full context of the issue is in the alert and the user can drill down to the trace in langfuse. What do you think? 

Furthermore, I love the force CLI feature to get stats on demand.

danchild: Regarding the per ticket email, would we use a JIRA api to send this email?

I'm all for the weekly status reports, but we need to think carefully about how we are managing data in redis before enabling this featre. My guess is that we will not keep all historical data in redis - but instead purge the "old" data when it no longer needed. Obviously, unfinished issues will never be purged. The question becomes how long do we keep finished issues and related data around. The design will affect how we think about this weekly report (or even a monthly or yearly report if we're ambitious). On a related note, we should consider how we will archive old data so that it isn't lost but instead put in cold storage (I would recommend a configurable automated process to do this, but also to have a CLI tool to trigger ad-hoc archives)

proposals/Jira-workflow-stats-reporting.md
```diff
@@ -0,0 +1,281 @@
...
+Teams manually correlate Jira comment history and Langfuse traces to recreate timelines. There is no rollup for weekly status. Someone has to manually aggregate tickets together. It's slow and prone to error. 
+
+## Proposal
+
+### Overview
+
+Four complementary features:
+
+1. **Per-ticket summary** — a Jira comment posted automatically when the workflow completes successfully, populated with information about iteration counts at each stage, stage durations, tokens used, links to PRs & CI attempts. When a workflow is blocked due to escalation to human, the full stats summary is always posted. If token usage also exceeds a configurable threshold (`FORGE_COST_ALERT_THRESHOLD`), a cost alert section is included.
+2. **On-demand stats** — users can request current stats for any feature/bug ticket via `/forge stats` (Jira comment command) or `forge-watch stats PROJ-123` (CLI). Returns the same summary format as the per-ticket summary, reflecting the current state of the workflow regardless of completion status.
```
danchild: by same summary format, do you mean same summary information, formatting that information in an appropriate and readable stdout format (e.g. ASCII tables)?

proposals/Jira-workflow-stats-reporting.md
```diff
@@ -0,0 +1,281 @@
...
+3. **Per-ticket email notification** — when a workflow completes, an email with the aggregate stats is sent to the Jira ticket's reporter and assignee.
+4. **Weekly status report** — aggregated view across all tickets per project with activity during the reporting period, delivered via admin CLI command.
+
+All of these features pull data from LangGraph's checkpoint state, which is augmented with statistic fields that are recorded during execution of the workflow.
```
danchild, line 39: Just adding on here: data is also pulled from Langfuse, and analytical insights would be generated from a combination of Langfuse's trace/observation data and LangGraphs checkpoint state.

proposals/Jira-workflow-stats-reporting.md
ekuris-redhat: Thanks Dan! Glad the terminal-state notification and CLI ideas resonate.

On the Jira field vs comment, I originally considered richer per-step updates, but after discussing with Ella we landed on posting the stats summary at two points: always on blocked-due-to-escalation (so the team sees exactly what Forge did before they need to intervene), and on successful completion. The rationale was to keep the signal focused , updating after every step would add noise to the ticket timeline, and most of the time people don't need to see partial stats unless something went wrong. The /forge stats command covers the case where someone wants to check in on a running workflow.
The scorecard and alerts ideas are interesting but I'd keep them out of scope here — this proposal is focused on getting the basic stats instrumentation and reporting in place first. Once we have the data flowing, we'll be in a much better position to layer on scorecards and alert integration as a follow-up. What do you think?

proposals/Jira-workflow-stats-reporting.md
ekuris-redhat: Yes exactly, same data, but formatted for the terminal. The Jira comment uses wiki markup (panels, tables), while the CLI would output something more like readable ASCII tables or plain text. The information is the same, the presentation adapts to the medium.

proposals/Jira-workflow-stats-reporting.md
ekuris-redhat: Good callout. Right now the proposal keeps things simple by pulling everything from checkpoint state ,that way we don't add a runtime dependency on the Langfuse API during reporting. But you're right that Langfuse has richer trace/observation data that could add depth to the analysis. For this first pass I'd keep it checkpoint-only for reliability, and we can layer in Langfuse queries later if we want deeper insights (e.g., per-call latency breakdowns, individual prompt performance). The Langfuse session link is already included so people can drill in manually.

proposals/Jira-workflow-stats-reporting.md
```diff
@@ -2,11 +2,11 @@
...
+When a Forge workflow reaches a terminal state — successful completion, blocked, or unrecoverable failure — the system should automatically post a development statistics summary as a Jira comment on the Feature ticket. The primary quality signal is iteration count: how many manual interventions and corrections were needed at each stage. When a workflow is blocked due to escalation to human, the full stats summary is posted so the team can assess the work Forge completed before hitting the block. If token consumption also exceeds a configurable threshold, a cost alert is included. Users can request stats on demand for any ticket via `/forge stats` (Jira comment) or `forge-watch stats PROJ-123` (CLI). A per-ticket email with aggregate stats is sent to the ticket's reporter and assignee on workflow completion. Additionally, a weekly status report aggregates statistics across all active and completed tickets per project, delivered via admin CLI.
```
eshulman2: 1. forge watch is the CLI for the poller and can't access any information regarding the ticket. so the forge-watch command is irrelevant especially having that in pro deployment we do not plan on having the poller as part of the deployment.
2. if you can get the same info for a ticket with a comment command in the ricket why do you need a command to send the same info to your email?
3. the weekly report thing is sent to the mailing list isn't it? for pre-project email would you like to send it to anyone in the project?

proposals/Jira-workflow-stats-reporting.md
```diff
@@ -30,12 +30,14 @@ Teams manually correlate Jira comment history and Langfuse traces to recreate ti
...
 
-1. **Per-ticket summary** — a Jira comment posted automatically when the workflow ends, populated with information about iteration counts at each stage, stage durations, tokens used, links to PRs & CI attempts.
-2. **Weekly status report** — report aggregating information about all tickets per project which saw activity during the reporting period. Report available via admin CLI or emailed to a mailing list for self-service subscription.
+1. **Per-ticket summary** — a Jira comment posted automatically when the workflow completes successfully, populated with information about iteration counts at each stage, stage durations, tokens used, links to PRs & CI attempts. When a workflow is blocked due to escalation to human, the full stats summary is always posted. If token usage also exceeds a configurable threshold (`FORGE_COST_ALERT_THRESHOLD`), a cost alert section is included.
+2. **On-demand stats** — users can request current stats for any feature/bug ticket via `/forge stats` (Jira comment command) or `forge-watch stats PROJ-123` (CLI). Returns the same summary format as the per-ticket summary, reflecting the current state of the workflow regardless of completion status.
```
eshulman2: same as mentioned before the forge watch is the poller cli not a part of a forge production deployment.

proposals/Jira-workflow-stats-reporting.md
```diff
@@ -30,12 +30,14 @@ Teams manually correlate Jira comment history and Langfuse traces to recreate ti
...
+3. **Per-ticket email notification** — when a workflow completes, an email with the aggregate stats is sent to the Jira ticket's reporter and assignee.
```
eshulman2: the user will anyway get an email update from jira itself as it adds a comment on the ticket with the info in the comment, adding this email would require support inside forge itself and not only in the cli as discussed.

proposals/Jira-workflow-stats-reporting.md
```diff
@@ -155,40 +162,43 @@ A new module `workflow/stats/weekly_report.py` that aggregates data across all c
 | Iteration Analysis | Most revised stage, CI first-pass rate, avg iterations per completed ticket |
 | Token Consumption | Totals and percentage breakdown by stage |
 
-#### 5. Weekly Report Delivery
+#### 5. Report Delivery
```
eshulman2, line 164: this means that for every project we again need to add a static config to the script

proposals/Jira-workflow-stats-reporting.md
```diff
@@ -155,40 +162,43 @@ A new module `workflow/stats/weekly_report.py` that aggregates data across all c
...
-```bash
-forge weekly-report --email                        # configured recipients
-forge weekly-report --email --to "mgr@company.com" # override recipients
-```
+Scheduling is configured by the admin — a cron job or CI pipeline invokes the CLI command on a weekly cadence.
 
-- Report sent to a configured mailing list address — team members self-subscribe to receive reports and browse history in their inbox
-- Email delivery via Gmail SMTP (`smtp.gmail.com`) or the Gmail API
-- HTML-formatted email matching the CLI report content
+**Per-ticket email — automatic on completion:**
```
eshulman2: rejecting as mentioned before.

proposals/Jira-workflow-stats-reporting.md
```diff
@@ -155,40 +162,43 @@ A new module `workflow/stats/weekly_report.py` that aggregates data across all c
...
+When a workflow completes successfully, the team sees a structured summary comment appear on the Jira ticket. The ticket's reporter and assignee receive an email with the same stats. When a workflow is blocked due to escalation to human, the full stats summary is posted and emailed — if the cost threshold is also exceeded, a cost alert is included.
+
+**On-demand stats — any time, any ticket:**
+
+```bash
+# Jira comment:
+/forge stats
+
+# CLI:
+$ forge-watch stats PROJ-123
```
eshulman2: won't work

proposals/Jira-workflow-stats-reporting.md
```diff
@@ -247,28 +253,24 @@ $ crontab -e
 3. **Phase 3: Per-ticket summary** — Implement `workflow/stats/summary.py`, wire into `aggregate_feature_status`, format Jira comment. (~0.5 day)
 4. **Phase 4: Weekly report engine** — Implement `workflow/stats/weekly_report.py` with checkpoint scanning, aggregation, and report formatting. (~1 day)
 5. **Phase 5: CLI command** — Add `forge weekly-report` subcommand with `--days`, `--output`, `--format` flags. (~0.5 day)
-6. **Phase 6: Email delivery** — Implement `integrations/email/client.py`, add `--email` flag, SMTP configuration, mailing list support. (~0.5 day)
+6. **Phase 6: Email delivery** — Implement `integrations/email/client.py`, wire per-ticket email on workflow completion and blocked-due-to-escalation, SMTP configuration, resolve recipients from Jira ticket reporter/assignee. (~0.5 day)
```
eshulman2: rejecting this

eshulman2 (changes requested):

proposals/Jira-workflow-stats-reporting.md
ekuris-redhat: You're right on all three points. 1) Removed forge-watch stats, it's the poller CLI and won't be part of production deployment. On-demand stats are now only via /forge stats Jira comment. 2) Dropped per-ticket email, Jira already notifies watchers/reporter/assignee when the summary comment is posted, so a separate email is redundant. 3) On the weekly report mailing list, you raise a fair point about static config per project. How would you prefer we handle weekly report delivery? Should we keep it as CLI-only output that admins share however they want, or is there a delivery approach you'd prefer?

proposals/Jira-workflow-stats-reporting.md
ekuris-redhat: Removed. On-demand stats are now only via /forge stats Jira comment.

proposals/Jira-workflow-stats-reporting.md
ekuris-redhat: Agreed, dropped it. Jira's own notifications already cover this when the summary comment is posted.

proposals/Jira-workflow-stats-reporting.md
ekuris-redhat: This ties into the question above about weekly report delivery. Happy to simplify this, what approach would work for you?

proposals/Jira-workflow-stats-reporting.md
ekuris-redhat: Removed from the proposal.

proposals/Jira-workflow-stats-reporting.md
ekuris-redhat: Removed from the proposal.

proposals/Jira-workflow-stats-reporting.md
ekuris-redhat: Removed. Phase 6 (email delivery) is dropped, tests are now Phase 6.

proposals/Jira-workflow-stats-reporting.md
ekuris-redhat: Updated the approach. The weekly report now uses a per-project "Forge Weekly Status Report" ticket (auto-created on first run, labeled forge:weekly-report). The report is posted as a comment on that ticket, and email is sent to stakeholders via Jira's notification API, recipients resolved dynamically from project roles (Administrator, Scrum Master, Project Lead). No static config per project needed.

proposals/Jira-workflow-stats-reporting.md
```diff
@@ -172,31 +171,31 @@ forge weekly-report --project PROJ --output report.md   # file export
 forge weekly-report --format json --output r.json       # JSON for tooling
 ```
 
-Scheduling is configured by the admin — a cron job or CI pipeline invokes the CLI command on a weekly cadence.
+**Email delivery via Jira:**
 
-**Per-ticket email — automatic on completion:**
+Each project gets a single long-lived "Forge Weekly Status Report" ticket (Task type, labeled `forge:weekly-report`). The CLI finds or creates this ticket automatically on first run. Each weekly report is posted as a new comment on that ticket, and the Jira notification API (`POST /issue/{key}/notify`) sends the report to stakeholders in the project's Administrator, Scrum Master, and Project Lead roles.
```
eshulman2: not sure where this came from but this was rejected before I believe.

proposals/Jira-workflow-stats-reporting.md
```diff
@@ -172,31 +171,31 @@ forge weekly-report --project PROJ --output report.md   # file export
...
-$ forge-watch stats PROJ-123
 ```
 
-Both return the current stats snapshot for the ticket regardless of workflow status.
+Post this as a Jira comment on any ticket to get the current stats snapshot regardless of workflow status.
+
+**Weekly report — admin CLI (on-demand):**
 
-**Weekly report — admin CLI (on-demand or scheduled):**
+The first run auto-creates a "Forge Weekly Status Report" ticket in the project. Stakeholders are notified via Jira's built-in email. All past reports are browsable as comments on that ticket.
```
eshulman2: there is no forge weekly report ticket

eshulman2 (changes requested):

proposals/Jira-workflow-stats-reporting.md
ekuris-redhat: I think you might be thinking of the per-ticket email, which was rejected and removed. This is different, this is the delivery mechanism for the weekly status report. Since Jira's notification API only works on issues, we need a ticket to anchor the notification. The CLI auto-creates a single "Forge Weekly Status Report" ticket per project on first run and uses it to post the report and notify stakeholders via Jira. No SMTP, no static config per project, recipients come from Jira project roles. Does that make sense, or would you prefer a different approach for delivering the weekly report?

proposals/Jira-workflow-stats-reporting.md
ekuris-redhat: Right, it doesn't exist yet. The idea is that the CLI creates this ticket automatically on the first run for each project. It's a one-time auto-created ticket that serves as the anchor for posting weekly reports and sending notifications via Jira's API, so we don't need any static config or SMTP setup. Each subsequent run reuses the same ticket and adds the new report as a comment.

PR #59, opened 2026-05-28
Feature/update docs v2

csoceanu:
## Summary

Adds support for updating documentation that lives in a separate repository from the code. When the `forge.docs_repo` Jira project property is set, Forge automatically updates the docs repo after a code PR is merged.

- **New node `update_docs_repo`** — post-merge node that clones both repos, runs the update-docs skill with the code repo mounted read-only, and creates a fork-based PR for the docs changes
- **New Jira property `forge.docs_repo`** — optional project property pointing to the docs repo (`owner/repo` format)
- **`extra_mounts` in ContainerRunner** — allows mounting additional read-only volumes into containers
- **Wired into both workflows** — bug: `post_merge_summary → update_docs_repo → END`, feature: `human_review_gate → update_docs_repo → complete_tasks`
- **Non-blocking** — if `forge.docs_repo` is not set or the update fails, the workflow proceeds normally
- **Improved update-docs skill** — broader file discovery, case-insensitive grep (`-rlFi`), two-pass evaluation, "review beyond grep" step, skip auto-generated files

## How it works

1. After a code PR is merged, `update_docs_repo` checks `forge.docs_repo` via Jira project properties
2. If set (and different from the code repo), clones the code repo's fork branch + the docs repo
3. Runs the update-docs skill in a container with both repos mounted (`/workspace` = docs, `/code-repo` = code read-only)
4. If the agent made changes: forks the docs repo, pushes, creates a PR, posts the link as a Jira comment
5. Cleans up both workspaces

## Test plan

- [x] 13 unit tests (routing, Jira client, container mounts) — all pass
- [x] 936 total unit tests — no regressions
- [x] Both graphs compile
- [x] Ruff clean


ekuris-redhat (commented):
Hey, CI is failing because update_docs_repo.py imports _clean_forge_gitignore
  from forge.workflow.nodes.implementation but that function doesn't exist
  there or anywhere in the codebase. Every test suite that touches either graph
  fails to collect because of this. You either need to define and export it
  from implementation.py, or drop this approach entirely and use
  .git/info/exclude the way workspace_setup.py:327 already does, that's the
  existing pattern for keeping .forge/ local without modifying tracked files.

  A few more things to fix while you're at it:

  state.get("fork_owner", "") on line 67 returns None when the stored state
  value is None, not ""  Python's .get() default only kicks in when the key is
  missing entirely. Both states default fork_owner and fork_repo to None, so
  this produces github.com/None/None.git and fails silently. Use
  state.get("fork_owner") or "" instead.

  On line 72, if docs_manager.create_workspace() raises, the inner try/finally
  at line 85 is never entered and code_workspace (already created on line 73)
  leaks. Restructure so cleanup is guaranteed regardless of which workspace
  creation fails.

  On line 171, docs_repo.split("/") will raise ValueError if someone configures
  a value like "org/sub/repo" , get_project_docs_repo only checks that a /
  exists, not that there's exactly one. Use split("/", 1) or tighten the
  validation in the Jira client.

src/forge/prompts/v1/update-docs-separate.md
```diff
@@ -0,0 +1,15 @@
+Update documentation files that have become stale due to code changes in a separate repository.
+
+## Workspace
+
+{workspace_path}
+
+The documentation repository is mounted at `/workspace` (read-write). The code repository is mounted at `/code-repo` (read-only).
+
+Run `cd /code-repo && git diff origin/main...HEAD` to see the code changes, and check `/code-repo/.forge/handoff.md` for context on what was implemented and why.
```
danchild: `git diff origin/main...HEAD` hardcodes the name of the default branch. We should pass the default branch into a template variable. Something like `{code_default_branch}`

src/forge/workflow/nodes/update_docs_repo.py
```diff
@@ -0,0 +1,232 @@
...
+            branch_name=branch_name,
+        )
+
+        # Clone upstream code repo; try to reach the PR branch so the agent
+        # can run `git diff origin/main...HEAD` inside the container.
+        code_git = GitOperations(code_workspace)
+        code_git.clone()
+        if fork_owner and fork_repo:
+            code_git.add_fork_remote(fork_owner, fork_repo)
+            code_git.checkout_branch(branch_name, remote="fork")
```
danchild: In the case where a feature branch is deleted following a GitHub merge, attempting to checkout a feature branch will fail

src/forge/prompts/v1/update-docs-separate.md
ekuris-redhat: Added a `code_default_branch` template variable to the prompt and fetch the code repo's default branch via GitHub API alongside the existing docs-repo fetch. The diff command in the prompt is now `git diff origin/{code_default_branch}...HEAD`.

src/forge/workflow/nodes/update_docs_repo.py
ekuris-redhat: Wrapped both checkout paths in a `try/except GitError`. When checkout fails (branch deleted after merge), we look up the PR's `merge_commit_sha` via the GitHub API using `current_pr_number` from state and do a detached-HEAD checkout of that commit instead. Added `GitOperations.checkout_commit(sha)` for this. If neither the branch nor the PR number is available, we log a warning and proceed from default branch HEAD.

src/forge/workflow/feature/state.py
```diff
@@ -40,6 +40,9 @@ class FeatureState(
     parallel_branch_id: int | None
     parallel_total_branches: int | None
 
+    # Documentation
+    docs_pr_url: str | None
```
danchild, line 44: optional `docs_pr_url` member should be initialized to `None` in `defaults` dict defined in `create_initial_feature_state`

src/forge/workflow/bug/state.py
```diff
@@ -25,6 +25,9 @@ class BugState(
     bug_fix_implemented: bool
     tdd_approach: bool
 
+    # Documentation
+    docs_pr_url: str | None
```
danchild, line 29: optional `docs_pr_url` member should be initialized to `None` in `defaults` dict defined in `create_initial_feature_state`

src/forge/workflow/feature/state.py
ekuris-redhat: Added "docs_pr_url": None to create_initial_feature_state defaults.

src/forge/workflow/bug/state.py
ekuris-redhat: Added "docs_pr_url": None to create_initial_bug_state defaults.

danchild:
@ekuris-redhat a lot of time has passed since this feature was requested - it needs to be implemented in the new task workflow

danchild:
@ekuris-redhat the design of this feature assumes that users will need or want to configure one docs repository for each JIRA project. Is this a limitation we are ok to accept for now?

ekuris-redhat:
> @ekuris-redhat a lot of time has passed since this feature was requested - it needs to be implemented in the new task workflow
Why do you think it should be implemented as a new task workflow and not a feature workflow?

ekuris-redhat:
> @ekuris-redhat the design of this feature assumes that users will need or want to configure one docs repository for each JIRA project. Is this a limitation we are ok to accept for now?

Yes, this is a deliberate design choice and the limitation is intentional for now. The assumption is that the vast majority of
setups will be one project → one docs repo, which mirrors how we designed forge.prd_proposals_repo.
The main scenario this doesn't cover is a project that needs to update multiple separate docs repos from one code change. If that becomes a real need, the extension is straightforward: change the property from a single string to a list and iterate. But I'd rather validate the simpler version first before adding that complexity.
Is there a specific case you have in mind where a single Jira project would need more than one docs repo?

danchild:
> > @ekuris-redhat a lot of time has passed since this feature was requested - it needs to be implemented in the new task workflow
> > Why do you think it should be implemented as a new task workflow and not a feature workflow?

Unless I'm mistaken, a task could change code in one repo which requires changing documentation in the `doc_repo`

danchild:
> > @ekuris-redhat the design of this feature assumes that users will need or want to configure one docs repository for each JIRA project. Is this a limitation we are ok to accept for now?
> 
> Yes, this is a deliberate design choice and the limitation is intentional for now. The assumption is that the vast majority of setups will be one project → one docs repo, which mirrors how we designed forge.prd_proposals_repo. The main scenario this doesn't cover is a project that needs to update multiple separate docs repos from one code change. If that becomes a real need, the extension is straightforward: change the property from a single string to a list and iterate. But I'd rather validate the simpler version first before adding that complexity. Is there a specific case you have in mind where a single Jira project would need more than one docs repo?

One repo works - I was just raising the issue as a potential limiting factor. Agreed that one repo would cover the grand majority of cases

danchild:
@ekuris-redhat the feature is working as designed. One last point to bring up before merging. Most, if it all ,forge workflows block to control behavior. If this is still the desired behavior, we are good to go merge. If not, a gate needs to be added after the docs PR is created.

ekuris-redhat:
Hey Dan, thanks for the thorough review.

On the task workflow point , you are right. A task takeover can absolutely introduce code changes that require documentation updates in the docs repo. I wired `update_docs_repo` into the task takeover graph so that after human review approval the flow is now `human_review_gate → update_docs_repo → complete_task_takeover`, consistent with how the feature and bug workflows handle it. Also added `docs_pr_url` to `TaskTakeoverState` and updated the routing tests.

On the gate question, non-blocking is intentional. The docs PR is a best-effort update, if the agent makes no changes or the update fails the workflow completes normally either way. Adding a gate here would mean a docs review could block a code change from being fully closed, which felt like the wrong tradeoff. If we find in practice that docs PRs need human sign-off before the ticket closes we can revisit, but I'd rather validate the simpler version first.
Please review again my latest changes

danchild:
> Hey Dan, thanks for the thorough review.
> 
> On the task workflow point , you are right. A task takeover can absolutely introduce code changes that require documentation updates in the docs repo. I wired `update_docs_repo` into the task takeover graph so that after human review approval the flow is now `human_review_gate → update_docs_repo → complete_task_takeover`, consistent with how the feature and bug workflows handle it. Also added `docs_pr_url` to `TaskTakeoverState` and updated the routing tests.
> 
> On the gate question, non-blocking is intentional. The docs PR is a best-effort update, if the agent makes no changes or the update fails the workflow completes normally either way. Adding a gate here would mean a docs review could block a code change from being fully closed, which felt like the wrong tradeoff. If we find in practice that docs PRs need human sign-off before the ticket closes we can revisit, but I'd rather validate the simpler version first. Please review again my latest changes

@ekuris-redhat the new addition looks good to me. It should work as intended

eshulman2 (changes requested):
## Requesting changes: the docs-repo PR needs to go through the same lifecycle as any other Forge-managed PR

**The ask:** the PR that `update_docs_repo` opens in `forge.docs_repo` should be treated like every other Forge-managed implementation PR — registered for tracking, subject to human review with a real revision loop, and reachable by the same gate/command surface (`/forge rebase`, `/forge skip-gate`, CI gating) — not opened and then abandoned to Forge.

**Why this is a blocker, not a nice-to-have:** today it isn't tracked at all. `_create_docs_pr()` only sets `docs_pr_url` (a string) — it never adds an entry to `state["pull_requests"]`, which is the structure `event_targets_pull_request()` (`src/forge/workflow/pr_state.py`) uses to route webhooks back into a running workflow. Concretely, as it stands:

- A reviewer requesting changes on the docs PR has no effect — `human_review_gate` / `review_response_gate` never see it, so there's no revision loop, unlike the human-review step every code PR goes through.
- `/forge rebase` and `/forge skip-gate` don't verify which repo/PR they're targeting (`worker.py::_handle_resume_event` keys off `current_state["current_pr_number"]` alone). Since the docs PR reuses the ticket key in its title/branch, these commands get routed to the ticket's workflow thread and act on `current_repo`/`current_pr_number` from state — the **original, already-merged code PR** — not the docs PR the human is actually looking at. The command silently misfires instead of doing what was asked.

Net result: the docs PR is dead on arrival. It looks like a normal Forge PR (ticket key in the title, same branch convention) but none of Forge's usual mechanics — review response, rebase, gating — actually work on it.

**What I'd like to see before this merges:** the docs PR gets the same treatment as the code PR it's paired with — added to `state["pull_requests"]`, wired through the review/CI gate loop so review feedback actually triggers a revision, and correctly targeted by `/forge rebase` / `/forge skip-gate`. If that's a larger change than fits here, let's at least scope it as an explicit fast-follow with an issue filed, rather than merging a PR path that looks Forge-managed but silently isn't.

PR #64, opened 2026-06-01
fix: preserve all PRD sections during revision

ekuris-redhat:
## Summary

- **Root cause**: `_strip_preamble` checked `text.find("\n#")` before `text.startswith("#")`. When the LLM revision output started with `## 1. Executive Summary` (no document title), the method found `\n## 2. Problem Statement` and stripped everything before it — including the entire Executive Summary.
- **Fix**: Swap the check order — `startswith("#")` first so content that already begins with a heading is never stripped.
- **Also**: Strengthened the shared `regenerate.md` prompt with explicit preservation rules for all content types (PRD, spec, epic, task).

## Changes (3 files)

| File | Change |
|------|--------|
| `src/forge/integrations/agents/agent.py` | `_strip_preamble`: check `startswith("#")` before `find("\n#")` |
| `src/forge/prompts/v1/regenerate.md` | Added explicit "Preserve what was not objected to" instructions |
| `tests/unit/prompts/test_prompt_templates.py` | Regression test for preservation instructions |

## Test plan

- [x] All 21 prompt template unit tests pass
- [x] E2E on AISOS-742 (no new skill, `_strip_preamble` fix only):
  - Generated PRD → rejected with "change threshold to 500K"
  - Executive Summary: **preserved**
  - All 11 sections: **intact**
  - Only requested change applied: 50,000 → 500,000 (3 references, 0 old values remain)


eshulman2:
fixed in #75

ekuris-redhat:
Closing this PR — we were unable to consistently reproduce the Executive Summary deletion during PRD revision in local end-to-end testing.

While we did observe `_strip_preamble` stripping the Executive Summary once (when the LLM output started directly with `## 1. Executive Summary` without a document title), the issue is non-deterministic and depends on the LLM's output formatting choices. Across ~15+ revision attempts, the bug only triggered once.

The root cause hypothesis (check order in `_strip_preamble` + weak regeneration prompt) remains plausible but needs more evidence before merging a fix. Will revisit if the issue is reported again with reproducible steps.

PR #75, opened 2026-06-07
feat: require ! prefix for revision comments, ignore unprefixed

eshulman2:
### Revision prefix for comments
Previously, any Jira comment that wasn't a question (\`?\` prefix) would trigger artifact regeneration — meaning users couldn't post informational comments ("looks good", "FYI for the team") without accidentally kicking off a revision cycle.

Now comments require an explicit \`!\` prefix to trigger revisions. Unprefixed comments are treated as informational and ignored by the workflow.

| Prefix | Effect |
|--------|--------|
| \`!\` | Revision request — triggers regeneration with feedback |
| \`?\` or \`@forge ask\` | Question — triggers Q&A answer |
| \`>option N\` | RCA option selection (RCA Option Gate only) |
| _(no prefix)_ | Informational — workflow ignores it |

### Fix: PRD section loss during revision
\`_strip_preamble\` used \`text.find("\n#")\` which matched any \`\n#\` anywhere — including \`#1\` in numbered lists or heading-like patterns in agent narration. This could strip actual document content (e.g., the Executive Summary section was being silently dropped after revisions).

Fixed by:
- Making \`_strip_preamble\` only match proper headings (\`# \`, \`## \`, \`### \` with a space) within the first 5 lines
- Updating the regenerate prompt to explicitly require preserving ALL sections from the original document

### Fix: forge:retry at approval gate regenerates instead of advancing
When \`forge:retry\` was applied at a paused approval gate with no prior error, the workflow was treating it as an approval signal and advancing to the next stage instead of regenerating the artifact.

Root cause: the gate's conditional edge (\`route_prd_approval\` etc.) routes to the approved path when \`is_paused=False\` and \`revision_requested=False\`. The previous attempt set \`revision_requested=False\` and tried to override routing by writing \`current_node\` directly into state — but LangGraph routing is driven by conditional edges, not by the \`current_node\` state field.

Fixed by setting \`revision_requested=True\` and \`feedback_comment\` so the gate's conditional edge correctly routes to the regeneration node on resume.

## Test plan
- [ ] Post \`!Fix the auth section\` on a paused approval gate — verify regeneration triggers
- [ ] Post \`?Why REST?\` on a paused gate — verify Q&A answer is posted
- [ ] Post \`Looks good, thanks\` on a paused gate — verify workflow stays paused (no regeneration)
- [ ] Post \`>option 2\` at RCA option gate — verify option selection still works
- [ ] Trigger a PRD revision — verify all sections (including Executive Summary) are preserved
- [ ] Apply \`forge:retry\` at a paused PRD/spec/plan gate with no error — verify regeneration, not approval
- [ ] Apply \`forge:retry\` at a failed node (with \`last_error\` set) — verify normal error-clear behaviour unchanged
- [ ] 932 unit tests pass


ekuris-redhat:
I've tested the PR and can confirm that I am reproducing the exact issue with the executive summary disappearing.
https://redhat.atlassian.net/browse/AISOS-886

ekuris-redhat:
LGTM 
Tested PR #75 locally against Jira tickets. Verified that the PRD executive summary preservation fix works as expected. Additionally, confirmed that the new behavior successfully ignores comments without requiring the ! prefix.

ekuris-redhat (approved):

PR #143, opened 2026-07-13
Add Concurrent CI Review Gate

### LINKED ISSUE #137: Run CI evaluation and human review concurrently instead of sequentially
## Problem

The current workflow forces CI and human review to be sequential: `wait_for_ci_gate → ci_evaluator → human_review_gate`. This causes two issues:

1. **Repos with no CI configured** get stuck forever at `wait_for_ci_gate` — no `check_run` webhook ever arrives, so the workflow never proceeds to review.
2. **Reviewers are blocked** — they can't start reviewing until all CI checks pass, even though GitHub allows CI and review to happen concurrently.

## Proposed Solution

After PR creation, go directly to `human_review_gate` and accept both CI and review webhooks concurrently:

- **CI webhook** → update `ci_status` in state, trigger CI fix if needed, but stay at the review gate
- **Review webhook** → trigger `implement_review` as normal
- **Merge webhook** → complete the workflow (GitHub branch protection ensures both CI and review pass before merge)

CI fix attempts still work — if CI fails, the fix agent pushes a commit and the workflow continues waiting for both new CI results and review.

This matches how GitHub PRs actually work and naturally solves the "no CI" problem — if there are no checks, the workflow just waits for review and merge.

## Impact

All workflows (feature, bug, task takeover) share the same post-PR lifecycle nodes, so this change benefits all of them.

## Current Workaround

Use `/forge skip-gate ci` on the PR to manually bypass the CI gate.

danchild:
## Summary

Reworks the post-PR lifecycle so CI evaluation and human review run concurrently. Previously, workflows blocked at `wait_for_ci_gate` until all CI checks passed before proceeding to human review. Now, after PR creation the workflow goes directly to `human_review_gate` and accepts both CI and review webhooks in parallel — matching GitHub's natural PR workflow where merge is the final gate.

Key changes:
- Eliminate the `wait_for_ci_gate` node; CI webhooks are routed to `human_review_gate` instead
- Add `pending_ci_event` state field for buffering CI webhooks at `human_review_gate`
- Extract shared post-PR lifecycle wiring into `post_pr.py` (reused by feature, bug, task_takeover graphs)
- Add CI failure attribution phase before attempting fixes
- Fix CI webhook routing to preserve `current_node` at `human_review_gate` (prevents premature override to `ci_evaluator`)

**This change is not backward compatible.** In-flight workflows checkpointed at `wait_for_ci_gate` will need to be manually restarted.

## Test plan
- [x] All existing unit tests pass (1919 passed)
- [x] Fast Tests & Coverage CI gate passes
- [x] Contract & Workflow Flow Tests pass
- [x] Integration Tests pass
- [x] E2E Smoke Tests pass
- [x] Lint passes
- [x] Container Build passes
- [x] CI webhook at `human_review_gate` correctly sets `pending_ci_event` and preserves `current_node`
- [x] CI webhook at `ci_evaluator` still routes directly (no `pending_ci_event`)
- [x] Multi-repo CI webhook activates correct PR record and sets `pending_ci_event`
- [x] Review gate resume event handling fully covered (12 tests re-added for review_response_gate and human_review_gate behaviors)

Closes #137


src/forge/workflow/nodes/implement_review.py
```diff
@@ -384,12 +384,7 @@ async def implement_review(state: WorkflowState) -> WorkflowState:
             decisions=accepted_decisions,
         )
 
-        # Only re-enter the CI gate if we actually pushed new commits; otherwise
-        # CI won't re-trigger and wait_for_ci_gate would block forever.
-        if contested_comments:
-            next_node = "review_response_gate"
-        else:
-            next_node = "wait_for_ci_gate" if unpushed else "human_review_gate"
+        next_node = "human_review_gate"
```
eshulman2: `next_node` is hardcoded to `"human_review_gate"` even when objections exist, contradicting this function's own docstring and the code above that still builds `contested_comments` and posts objections.

**Failure scenario:** the analysis container marks a review comment as "contest"; Forge posts the objection to the PR/Jira as designed, but the workflow proceeds straight to `human_review_gate` instead of pausing at `review_response_gate`. Combined with the deleted inline-reply handler in `worker.py` (see review summary), a human replying to the contested thread has no mechanism to resolve it.

src/forge/workflow/nodes/human_review.py
```diff
@@ -7,33 +7,56 @@
...
+                    "🚀 Pull request created and submitted. Waiting for CI checks and human review."
+                )
+            await post_status_comment(jira, ticket_key, message)
+            await remove_implementing_label(jira, ticket_key)
+            await set_ci_pending_label(jira, ticket_key)
+        finally:
+            await jira.close()
+        logger.info(f"Pausing {ticket_key} at human_review_gate after PR creation")
+    else:
+        logger.info(f"Pausing {ticket_key} at human_review_gate (ci_status={ci_status!r})")
```
eshulman2, line 63: `human_review_gate` now unconditionally sets `is_paused: True`, removing the prior explicit `is_paused: False` branch for `pr_merged`; nothing downstream resets it.

**Failure scenario:** PR merges and routing correctly proceeds to `current_node="complete"` within the same `ainvoke`, but `is_paused` stays `True` (inherited via `{**state}` spreads). In `worker.py`, the completion-recording branch guarded by `if not is_paused` never fires for a normal completed workflow, and a later stray webhook logs "Workflow is paused at complete" and sets `should_resume=True` based on stale state.

src/forge/orchestrator/worker.py
```diff
@@ -603,6 +556,7 @@ async def _handle_resume_event(
 
         # GitHub issue_comment events: detect /forge skip-gate and /forge unskip-gate
         # commands posted as PR comments.
+        _CI_STAGES = ("ci_evaluator", "attempt_ci_fix", "human_review_gate")
```
eshulman2: A new local `_CI_STAGES` is defined inside `_handle_resume_event`, shadowing the module-level `_CI_STAGES` which still references the deleted `wait_for_ci_gate` node and omits `human_review_gate`.

**Failure scenario:** the module-level `_CI_STAGES` is now dead code diverging from reality. Any future refactor that removes the local shadow (e.g. extracting the skip-gate block into a helper) will silently fall back to the stale tuple, re-breaking `/forge skip-gate` at `human_review_gate`.

eshulman2 (commented):
Automated review: found several correctness regressions in the CI/human-review lifecycle rework that should be addressed before merge. Two additional issues fall outside this diff's changed lines and couldn't be attached inline, noting them here:

**`src/forge/orchestrator/worker.py` (around line 886, in `_handle_resume_event`)** — The dedicated handler that resumed the workflow when a human replied inline to a specific contested-comment thread at `review_response_gate` (matching `in_reply_to_id` against `contested_comments` and clearing only that thread) was deleted with no replacement; only the unrelated PRD/spec-proposal-PR inline-reply path remains (`is_proposal_reply` check). *Failure scenario:* a human posts an inline reply on a GitHub review thread that Forge previously contested on an implementation PR. No handling matches this event for `review_response_gate` anymore, so `is_rejected`/`is_approved` are never set and the workflow isn't resumed — the reply is silently ignored. Two tests exercising exactly this behavior (`test_inline_reply_resumes_only_its_contested_thread`, `test_standalone_inline_comment_is_actionable_at_response_gate`) were deleted rather than adapted.

**`src/forge/workflow/pr_state.py` (line ~100, `save_active_pull_request` / `_PR_LIFECYCLE_NODES`)** — This file wasn't touched by the diff but is broken by it: `save_active_pull_request` still defaults a new PR record's `lifecycle_node` to the now-deleted `"wait_for_ci_gate"` node, and `_PR_LIFECYCLE_NODES` still lists it. *Failure scenario:* in a multi-repo ticket, `create_pull_request` calls `save_active_pull_request` with `current_node="teardown_workspace"`, which isn't in `_PR_LIFECYCLE_NODES`, so `lifecycle_node` gets stamped as `"wait_for_ci_gate"` for every non-final repo. A later `/forge skip-gate <name>` comment targeting an earlier repo restores that dead node, which is no longer in `worker.py`'s local `_CI_STAGES` tuple, so the skip-gate command is silently dropped even though the ticket is genuinely at a CI-related stage.

eshulman2 (commented):
Before merging, please provide one of the following for workflows currently checkpointed in the CI stage:

1. Backward compatibility: retain a temporary route/alias from  to , including migration of any  values that still reference ; or
2. A clear, tested migration path for affected tickets that preserves their existing PR mappings, per-PR CI/review state, and branches.

Simply sending another webhook will resume the existing checkpoint rather than restart it, while clearing the checkpoint can discard PR lifecycle state and cause work to be regenerated. The PR description currently says workflows must be manually restarted, but does not define a safe procedure or provide tooling to identify and migrate affected tickets.

Please document the exact operator steps, expected state transformations, rollback behavior, and tests covering an in-flight workflow at  (including a multi-repository ticket).

eshulman2 (commented):
Correction/clarification to my previous review comment:

Before merging, please provide one of the following for workflows currently checkpointed in the CI stage:

1. Backward compatibility: retain a temporary route or alias from wait_for_ci_gate to human_review_gate, including migration of any per-PR lifecycle_node values that still reference wait_for_ci_gate; or
2. A clear, tested migration path for affected tickets that preserves their existing PR mappings, per-PR CI/review state, and branches.

Simply sending another webhook resumes the existing checkpoint rather than restarting it, while clearing the checkpoint can discard PR lifecycle state and cause work to be regenerated. The PR description currently says workflows must be manually restarted, but does not define a safe procedure or provide tooling to identify and migrate affected tickets.

Please document the exact operator steps, expected state transformations, rollback behavior, and tests covering an in-flight workflow at wait_for_ci_gate, including a multi-repository ticket.

src/forge/workflow/nodes/implement_review.py
danchild: valid

src/forge/orchestrator/worker.py
danchild: valid

src/forge/workflow/nodes/human_review.py
danchild: valid

eshulman2:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

PR #149, opened 2026-07-14
fix: fail closed when review analysis omits its plan

eshulman2:
## Summary

- require review analysis to produce a non-empty review-plan artifact
- treat a missing or empty plan as an implement_review failure so normal retry/error handling applies
- preserve the explicit no-action marker as the only successful no-op
- add regression coverage for missing and empty plans

## Why

During AISOS-2157, the analysis agent correctly analyzed all review feedback but ended before writing the required plan. Forge interpreted the missing file as no actionable feedback and silently returned to the human-review gate.

## Validation

Focused Ruff checks pass. All 26 implement-review unit tests pass.


eshulman2:
Closing this narrow fix for now. Robustly distinguishing missing analysis output from an intentional no-op requires a shared typed artifact contract rather than parsing sentinel text or Markdown structure. Tracking the broader design and migration in #150.

PR #155, opened 2026-07-15
fix: notify Jira after queue retries exhaust

### LINKED ISSUE #140: tech debt: notify Jira when workflow execution exhausts queue retries
## Context

PR #58 centralizes Jira notifications for handled workflow failures represented by `last_error`. A separate visibility gap remains when `compiled_workflow.ainvoke()` raises instead of returning workflow state: the worker logs and re-raises the exception, and the queue retries the event, but Jira is never notified if all retries fail.

This is intentionally out of scope for PR #58 so handled workflow errors and queue-level execution failures remain separate concerns.

## Proposed behavior

- Do not notify Jira for each retryable queue failure.
- When the existing retry mechanism exhausts its attempts and moves the event to the dead-letter queue, emit one terminal Jira notification.
- Keep the existing queue retry behavior unchanged.
- Include the ticket key, sanitized error summary, event/correlation ID, and recovery instructions.
- Use the shared error formatting/redaction path where practical.
- Add idempotency protection keyed by the event ID so terminal processing cannot create duplicate comments.
- A Jira notification failure must not prevent the message from reaching the dead-letter queue.

## Suggested design

Add an optional terminal-failure callback or event to `QueueConsumer`. Invoke it when `RetryQueue.enqueue_for_retry()` reports that the message was moved to the dead-letter queue, in both the initial consumer failure path and the retry poller. The orchestrator can provide the Jira-specific handler without coupling the generic retry queue directly to Jira.

## Acceptance criteria

- Retryable escaped workflow exceptions do not post Jira comments.
- Exhausting queue retries posts exactly one terminal Jira comment.
- The error text is secret-redacted.
- Duplicate terminal handling for the same event does not create another comment.
- Notification failure does not interfere with dead-lettering.
- Both the initial processing path and retry-poller path are covered by tests.

Related: #58

iamdhrv:
## Summary

- add an optional terminal-failure callback to `QueueConsumer` and invoke it only after a message reaches the dead-letter queue
- claim terminal handling atomically in Redis by event ID so multiple retry pollers cannot emit duplicate notifications
- have the orchestrator post a secret-redacted Jira error comment with the ticket, event/correlation ID, and recovery instructions
- check existing Jira comments for the event marker to preserve idempotency across ambiguous API failures or replays
- contain callback and idempotency-store failures so dead-letter handling and stream acknowledgement continue

Closes #140

## Test plan

- [x] `uv run pytest tests/unit/queue/test_consumer_retry.py tests/unit/orchestrator/test_worker.py -q` (55 passed)
- [x] `uv run pytest tests/unit/ -q` (1462 passed, 2 skipped)
- [x] `uv run ruff check src/forge/queue/consumer.py src/forge/orchestrator/worker.py tests/unit/queue/test_consumer_retry.py tests/unit/orchestrator/test_worker.py`
- [x] `uv run ruff format --check src/forge/queue/consumer.py src/forge/orchestrator/worker.py tests/unit/queue/test_consumer_retry.py tests/unit/orchestrator/test_worker.py`

## Type checking

`uv run mypy src/forge/queue/consumer.py src/forge/orchestrator/worker.py` still reports the existing strict-mode errors in these modules (15 total); the new callback and Jira handler add no new reported mypy errors.


src/forge/queue/consumer.py
```diff
@@ -268,7 +324,11 @@ async def _process_retry_queue(self) -> None:
                         # the counter keeps accumulating and the message can eventually
                         # reach the dead-letter queue.
                         await self._retry_queue.remove_from_retry_without_counter_reset(entry)
-                        await self._retry_queue.enqueue_for_retry(entry.message, str(e))
+                        queued_for_retry = await self._retry_queue.enqueue_for_retry(
+                            entry.message, str(e)
+                        )
+                        if not queued_for_retry:
```
eshulman2: Nice job keeping notification failures from interrupting DLQ handling. Could we also acknowledge the original stream entry when `queued_for_retry` is false? This branch currently notifies and then `continue`s, so terminal failures reached through the retry poller remain indefinitely in the Redis pending-entry list. A best-effort `xack`, regardless of whether the callback succeeds, plus a test for this path would keep the PEL clean.

src/forge/queue/consumer.py
```diff
@@ -134,6 +146,49 @@ async def _ack(self, stream: str, message_id: str) -> None:
         redis_client = await self._get_redis()
         await redis_client.xack(stream, CONSUMER_GROUP, message_id)
 
+    async def _notify_terminal_failure(self, message: QueueMessage, error: str) -> None:
```
eshulman2: Thanks for adding the cross-worker notification claim here. One concern: this claims only the callback; the retry entry itself is still read non-atomically by every worker via `get_due_messages()`, so two pollers can process the same event concurrently. If one succeeds while another exhausts the counter, we could move a successfully processed event to the DLQ and post a misleading terminal-failure comment; concurrent failures can also advance the retry counter more quickly than intended. Could we atomically claim/lease the retry entry before dispatching it (for example, with a Redis transaction or Lua operation), and add a multi-worker test?

iamdhrv: Addressed in 3f3036b. Due retry entries are now atomically claimed with a Redis Lua operation that advances their score to a 15-minute visibility lease before dispatch. A crashed worker’s entry becomes claimable again after the lease. Added both a concurrent-poller unit test and a real-Redis integration test; the full unit suite and PR CI pass.

src/forge/queue/consumer.py
iamdhrv: Addressed in 3f3036b. The retry-poller terminal path now performs a best-effort xack after the callback attempt, including when the callback fails. Added focused tests for both callback success and failure; the full unit suite and PR CI pass.

eshulman2 (changes requested):
Thanks for tackling this—moving terminal failures to the DLQ, containing callback failures, and adding cross-worker deduplication are useful improvements. I found three reliability issues that should be addressed before merge:

1. **Renew the retry-processing lease while the handler runs.**

`RETRY_CLAIM_LEASE_SECONDS` is 15 minutes, while Forge processing can exceed that (the container timeout alone defaults to 30 minutes). Once the score becomes due again, another worker can claim and execute the same retry concurrently.

Please add ownership-aware lease renewal while `_process_message(..., raise_on_error=True)` is running. A Redis Lua operation can conditionally extend the sorted-set score when the member still exists; renew substantially before expiry (for example, every one-third of the lease). Add a regression test where processing outlives the initial lease and verify a second poller cannot claim it.

2. **Make terminal-notification retries durable.**

The five-minute claim TTL does not currently provide recovery: after the callback fails or the worker dies, the event has already entered the DLQ and is acknowledged, and nothing schedules another notification attempt when the claim expires.

Please persist notification work independently from the stream message. One suitable design is a Redis sorted-set outbox created atomically with the DLQ entry: workers lease due notifications, invoke the callback, and remove an outbox entry only after successful delivery. A callback/Jira failure should leave it available after the lease expires. If a DLQ entry is manually requeued, remove its stale pending notification. Tests should cover callback failure followed by a later successful attempt and a worker-safe single claim.

3. **Paginate the Jira comment lookup used for idempotency.**

`JiraClient.get_comments()` performs one request even though Jira paginates comments. On tickets with enough comments, the event marker may be outside the returned page, allowing a duplicate notification after an ambiguous Jira response or replay.

Please follow `startAt`, `maxResults`, and `total` until all pages have been read, and add a test where the matching marker is on a later page.

The existing secret redaction, event marker, and best-effort acknowledgement behavior can remain. Happy to review the next revision.

PR #198, opened 2026-07-23
fix: skip rebase when remote branch does not exist yet

### LINKED ISSUE #199: Bug: pull_rebase fails when remote branch does not exist yet
## Summary

`pull_rebase()` attempts `git rebase origin/<branch>` unconditionally, but on first implementation runs the branch was just created locally and hasn't been pushed yet. The remote ref doesn't exist, so the rebase fails. Since this error is not recoverable (same failure on every retry), the workflow exhausts all 3 retries and enters `forge:blocked` state immediately after workspace setup.

Observed on AISOS-2275 and likely other tickets during first implementation.

## Root Cause

`workspace_setup` creates the branch locally from `origin/main` but never pushes it. The next node (`implement_task`) calls `prepare_workspace` → `pull_rebase("origin")`, which tries `git rebase origin/<branch>` — a ref that doesn't exist. The error is not recoverable (same failure on every retry), so the workflow blocks after 3 attempts.

## Fix

`pull_rebase()` should check `remote_branch_exists()` before attempting `git rebase origin/<branch>`. If the remote branch doesn't exist yet, skip the rebase — a fetch-only sync is sufficient since there are no remote changes to integrate.

danchild:
Fixes #199

## Summary
- `pull_rebase()` now checks `remote_branch_exists()` before attempting `git rebase origin/`, preventing failures on first implementation runs where the branch was just created locally and hasn't been pushed yet
- This was causing AISOS-2275 (and likely other tickets) to exhaust all 3 retries and enter `forge:blocked` state immediately after workspace setup

## Root Cause
`workspace_setup` creates the branch locally from `origin/main` but never pushes it. The next node (`implement_task`) calls `prepare_workspace` → `pull_rebase("origin")`, which tries `git rebase origin/` — a ref that doesn't exist. The error is not recoverable (same failure on every retry), so the workflow blocks after 3 attempts.

## Test plan
- [x] Added `test_pull_rebase_skips_when_remote_branch_missing` — verifies only `fetch` is called when remote branch is absent
- [x] Added `test_pull_rebase_rebases_when_remote_branch_exists` — verifies fetch + rebase both happen when remote branch exists
- [x] Existing `test_no_fork_state_sync_failure_raises` continues to pass


eshulman2:
duplicates #205

PR #227, opened 2026-07-27
Use forks for PRD and spec proposal pull requests

eshulman2:
## Summary

- create and synchronize a Forge-owned fork before publishing PRD and spec proposals
- write proposal branches and files to the fork, then open cross-repository PRs against upstream
- persist fork coordinates so feedback revisions update the fork while comments remain on the upstream PR
- retain fallback behavior for proposal PR state created before this change

## Testing

- `UV_CACHE_DIR=/tmp/uv-cache uv run pytest tests/unit/workflow/nodes/test_prd_pr.py tests/unit/workflow/nodes/test_spec_pr.py tests/unit/workflow/feature/test_prd_pr_state.py` (12 passed)
- focused Ruff checks passed
- `git diff --check` passed


eranco74 (commented):
Overall the approach is correct — fork-based workflow cleanly sidesteps the upstream write-permission issue. Two things to address before merging.

src/forge/workflow/nodes/prd_generation.py
```diff
@@ -83,10 +83,14 @@ async def _create_prd_proposal_pr(
     gh = GitHubClient()
     jira = JiraClient()
     try:
-        await gh.create_branch(owner, repo, branch)
+        fork = await gh.get_or_create_fork(owner, repo)
+        fork_owner = fork["owner"]["login"]
+        fork_repo = fork["name"]
```
eranco74: `sync_fork_with_upstream` return value is silently ignored here (and in `spec_generation.py`). If the fork has diverged (returns `False`), the code proceeds to create a branch from stale fork state — the resulting PR could have merge conflicts or be based on outdated `main`.

At minimum log a warning; ideally raise so the workflow surfaces the problem instead of creating a silently broken PR.

tests/unit/workflow/nodes/test_prd_pr.py
```diff
@@ -140,6 +155,8 @@ async def test_updates_file_on_branch(self):
             ticket_type=TicketType.FEATURE,
```
eranco74, line 194: The update test always sets `prd_pr_fork_owner` / `prd_pr_fork_repo`. Could you add a test where those fields are `None` to cover the backwards-compatible fallback path? That's the code path all existing in-flight workflows will hit immediately after deploy.

Same applies to `test_spec_pr.py`.

eranco74 (approved):

PR #229, opened 2026-07-27
fix: skip re-pause in review gates when PR is already merged

### LINKED ISSUE #228: human_review_gate unconditionally re-pauses workflow when PR is already merged
## Summary

`human_review_gate` in `src/forge/workflow/nodes/human_review.py:16-30` unconditionally calls `set_paused()` without checking `pr_merged`. When a PR merge event arrives at this gate, the worker correctly sets `is_paused=False` and `pr_merged=True`, but the gate node immediately re-pauses the workflow. The downstream router `route_human_review` never reaches its `pr_merged` check (line 48) because `is_paused` is checked first (line 53) and returns `END`.

## Checkpoint trace showing the bug

```
12:39:10.865163 | human_review_gate | paused=true  | merged=       ← merge event arrives
12:39:10.865903 | human_review_gate | paused=false | merged=true   ← worker correctly unpauses
12:39:10.871688 | human_review_gate | paused=false | merged=true   ← workflow invoked
12:39:10.874852 | human_review_gate | paused=true  | merged=true   ← gate re-pauses! BUG
12:39:11.257419 | human_review_gate | paused=true  | merged=true   ← stuck forever
```

## Impact

After a PR is merged, the workflow gets permanently stuck at `human_review_gate`. Jira ticket status is never transitioned to Done. Subtasks are never completed. This affects every ticket that goes through the PR merge flow.

## Root Cause

The gate node does not check `pr_merged` before pausing:

```python
def human_review_gate(state: WorkflowState) -> WorkflowState:
    return set_paused(state, "human_review_gate")  # always pauses
```

## Fix

Skip pausing when the PR is already merged:

```python
def human_review_gate(state: WorkflowState) -> WorkflowState:
    if state.get("pr_merged"):
        return update_state_timestamp({**state, "current_node": "human_review_gate"})
    return set_paused(state, "human_review_gate")
```

The same pattern should be checked in `review_response_gate` and any other gate that can receive a merge event while paused.

ekuris-redhat:
## Summary

- `human_review_gate` and `review_response_gate` unconditionally re-paused the workflow even when `pr_merged=True`, causing the post-merge completion flow to never execute
- Both gates now check `pr_merged` and skip pausing when the PR is already merged

Fixes #228

## Test plan

- [x] New test: `test_human_review_gate_skips_pause_when_pr_merged` — verifies gate does not re-pause when `pr_merged=True`
- [x] New test: `test_human_review_gate_pauses_when_pr_not_merged` — verifies normal pause behavior is preserved
- [x] New test: `test_review_response_gate_skips_pause_when_pr_merged` — same check for the review response gate
- [x] All 1753 existing tests pass with no regressions


eshulman2:
The fix looks correct and I do not see a blocker. One non-blocking hardening suggestion: in both `pr_merged` branches, explicitly return `is_paused=False` rather than preserving the incoming value. That would also recover older or manually patched checkpoints that contain both `pr_merged=True` and `is_paused=True` (particularly important for `review_response_gate`, whose router checks `is_paused` first).

It would also be useful to add a graph-level regression test proving that a merge resumes from each review gate and reaches the completion path; the current tests validate the gate functions in isolation.

ekuris-redhat:
Both items addressed in 3b8be01:

1. **`is_paused=False` hardening** — both `human_review_gate` and `review_response_gate` now explicitly set `is_paused=False` in the `pr_merged` branch, so stale checkpoints with both flags set won't get stuck at `END` in the router.

2. **Gate→router integration tests** — two new tests feed each gate's output directly into its router with `is_paused=True` + `pr_merged=True` and assert the flow reaches the completion path (not `END`). Also added dedicated stale-checkpoint unit tests for both gates.

PR #234, opened 2026-07-29
Handle GitHub review feedback per thread

### LINKED ISSUE #233: Handle GitHub review feedback per thread across proposal and implementation PRs
## Problem
Forge currently flattens all unresolved GitHub review comments into one feedback document. It loses thread identity, posts objections as top-level comments, and pauses the entire review when any one item is contested. This prevents Forge from implementing fair change requests while replying to or escalating only the disputed threads.

The same limitation affects generated PRD/specification proposal PRs and normal implementation PRs.

## Desired behavior
- Preserve GitHub review thread and comment identity, file location, author, resolution state, and reviewed commit.
- Classify each review item independently as accept, reply/contest, uncertain/clarify, stale, or duplicate.
- Implement accepted items even when other threads are contested.
- Reply to addressed, contested, and clarification items in their originating GitHub threads.
- Keep unresolved contested threads pending without blocking unrelated code or document changes.
- Re-fetch thread state after pushes and ignore resolved, outdated, or already-processed feedback.
- Use top-level PR comments only for overall progress summaries.
- Apply the same mechanism to PRD, specification, and implementation PR review flows.

## Acceptance criteria
1. A mixed review with accepted and contested threads results in accepted changes being pushed and a reply being posted only to each contested thread.
2. Human responses resume the corresponding thread rather than clearing every contested item.
3. PRD/spec revisions aggregate only accepted feedback and reply in-place to non-actionable threads.
4. Normal PR changes aggregate only accepted feedback and do not pause before implementing them.
5. Resolved, outdated, duplicate, and previously processed threads do not trigger repeat work.
6. State remains compatible with workflows checkpointed before this change.
7. Unit and workflow tests cover mixed reviews, thread replies, partial progress, and resume behavior.

eshulman2:
Closes #233

## Summary
- preserve unresolved GitHub review thread, comment, location, author, and commit identity
- add in-thread reply support and inline-reply webhook handling
- classify normal PR feedback per thread as accept, contest, clarify, or ignore
- implement accepted normal-PR feedback even when other threads remain contested
- classify PRD/spec review threads independently and revise only accepted or uncertain items
- reply in place to contested, ignored, and addressed proposal feedback
- retain processed thread decisions to avoid duplicate work and resume only the matching thread
- keep legacy checkpoint and REST fallback behavior compatible

## Validation
- Ruff passed on all changed Python files
- 90 affected-path unit, webhook, contract, worker, proposal, and review tests passed

The unrelated TEST_GATES_ROLLOUT_PLAN.md file was not included.


src/forge/workflow/nodes/implement_review.py
```diff
@@ -80,19 +109,81 @@ async def _fetch_pr_review_comments(owner: str, repo: str, pr_number: int, revie
...
+    return [
+        item
+        for item in data
+        if isinstance(item, dict)
+        and item.get("disposition") in allowed
+        and isinstance(item.get("thread_id"), str)
+    ]
+
+
+def _merge_review_decisions(
```
eranco74: The decision-merge pattern (dict keyed by `thread_id`, then `.update()`) also appears twice in `worker.py` (lines ~1271 and ~1548). Not a blocker, but if any of the three copies drift you'd get inconsistent merge behavior. Could consolidate into this helper and import it from the worker.

src/forge/prompts/v1/triage-proposal-review-threads.md
```diff
@@ -0,0 +1,33 @@
...
+- Never let one replied-to or ignored thread prevent accepted changes in other threads.
+- Preserve thread_id and comment_id exactly.
+
+<artifact>
+{artifact_content}
+</artifact>
+
+<threads>
+{review_threads}
+</threads>
```
eranco74, line 33: The classification quality here (`accept` vs `reply` vs `uncertain`) is the make-or-break for the whole feature. Might be worth monitoring early decisions in production to catch systematic misclassification — e.g. valid feedback triaged as `reply` would silently drop reviewer intent.

eranco74 (commented):
This is a big step forward for the feedback process — per-thread triage and in-thread replies are exactly the right model. Reviewers will actually know what happened with each comment instead of guessing. A couple of minor observations inline.

src/forge/orchestrator/worker.py
```diff
@@ -471,9 +490,56 @@ async def _handle_resume_event(
...
+            reply = payload.get("comment", {})
+            replied_to = reply.get("in_reply_to_id")
+            sender_login = payload.get("sender", {}).get("login", "")
+            forge_login = await self._get_forge_github_login()
+            if sender_login and sender_login == forge_login:
+                logger.debug("Ignoring Forge's own inline review comment")
+                return current_state
+            if replied_to:
+                contested = current_state.get("contested_comments", [])
+                remaining = [item for item in contested if item.get("comment_id") != replied_to]
```
ekuris-redhat: in_reply_to_id here is the ID of whatever comment the user hit "Reply" on.
  But Forge posts its own reply into contested threads (via _reply_to_review_threads in implement_review). When the reviewer responds to Forge's message, which is the natural flow in_reply_to_id will be Forge's comment ID, not the stored comment_id in contested_comments. The filter here won't match anything, so the contested thread is never cleared. The feature's core promise ("human response resumes only the matching thread") breaks in the common case where Forge has already replied.
Needs either: Forge reply IDs stored alongside each contested item so they can also be matched here, or a note explaining why this case is intentionally not handled.

src/forge/workflow/utils/proposal_review_threads.py
```diff
@@ -0,0 +1,121 @@
...
+
+    decisions: dict[str, dict[str, Any]] = {}
+    for item in parsed:
+        if not isinstance(item, dict) or item.get("thread_id") not in expected:
+            continue
+        disposition = item.get("disposition")
+        if disposition not in _DISPOSITIONS:
+            continue
+        thread = expected[item["thread_id"]]
+        comment_id = thread["comments"][-1].get("comment_id")
```
ekuris-redhat, line 44: thread["comments"][-1] crashes with IndexError if comments is empty. get_pull_request_review_threads filters empty threads, but this function is a standalone parser ,it accepts arbitrary input. The inline reply stub built in worker.py around line 900 constructs its own thread dict, and callers outside the GraphQL path aren't guaranteed to be non-empty. Same issue at line 51 in the fallback branch.

src/forge/orchestrator/worker.py
```diff
@@ -850,9 +989,17 @@ async def _handle_resume_event(
                         _owner, _repo = repo_full.split("/", 1)
                         gh = GitHubClient()
                         try:
-                            inline_comments = await gh.get_review_comments(
-                                _owner, _repo, pr_number, review_id
+                            proposal_review_threads = await gh.get_pull_request_review_threads(
+                                _owner, _repo, pr_number
                             )
+                            inline_comments = [
```
ekuris-redhat: Identical 8-line list comprehension, copy-pasted exactly. If the shape of a thread changes (e.g. comments[-1] becomes comments[0] for something), one copy will get fixed and the other won't. Should be extracted to a one-liner helper before this PR merges.

src/forge/workflow/nodes/implement_review.py
```diff
@@ -181,9 +277,25 @@ async def implement_review(state: WorkflowState) -> WorkflowState:
...
+            repo=_repo,
+            pr_number=pr_number,
+            decisions=response_decisions,
+        )
+
+        # Backward-compatible fallback if an older analysis prompt writes only
+        # the legacy objections file. New analysis never blocks accepted work.
         objections_path = Path(workspace_path) / _REVIEW_OBJECTIONS_FILE
-        if objections_path.exists():
+        if not decisions and objections_path.exists():
```
ekuris-redhat, line 279: The not decisions and objections_path.exists() branch is the only thing protecting workflows already checkpointed in flight at deploy time. Acceptance criterion 6 explicitly calls this out. There's no test that sets decisions = [], writes a review-objections.md, and asserts contested_comments and current_node are set correctly. This path could silently break on any future refactor.

src/forge/workflow/nodes/implement_review.py
```diff
@@ -152,6 +243,11 @@ async def implement_review(state: WorkflowState) -> WorkflowState:
             repo=_repo,
```
ekuris-redhat, line 224: The mechanism for acceptance criterion 5 ("previously processed threads don't trigger repeat work") is processed_thread_ids derived from state["review_comments"]. There's no test that runs implement_review twice on the same state, verifies review_comments accumulates from the first run, and confirms those thread IDs are excluded from the fetched comment text on the second run. Without this, the deduplication feature is untested.

src/forge/workflow/nodes/implement_review.py
```diff
@@ -80,19 +109,81 @@ async def _fetch_pr_review_comments(owner: str, repo: str, pr_number: int, revie
...
+    merged = {
+        item["thread_id"]: item
+        for item in previous
+        if isinstance(item, dict) and item.get("thread_id")
+    }
+    merged.update({item["thread_id"]: item for item in current})
+    return list(merged.values())
+
+
+async def _reply_to_review_threads(
```
ekuris-redhat, line 157: Same loop pattern: iterate decisions, skip if comment_id is not an int or
  response is empty, call github.reply_to_review_comment, swallow per-comment
  exceptions. The only differences are the status == "addressed" skip and the
  dispositions filter, both one-liners. Consider consolidating into one shared
  function so GitHub reply behavior stays consistent.

src/forge/workflow/utils/proposal_review_threads.py
```diff
@@ -0,0 +1,121 @@
...
+            context={"ticket_key": ticket_key},
+            include_tools=False,
+        )
+    except Exception as exc:
+        logger.warning("Proposal thread triage failed for %s: %s", ticket_key, exc)
+        output = ""
+    return parse_proposal_thread_decisions(output, threads)
+
+
+async def reply_to_proposal_decisions(
```
ekuris-redhat, line 105: Same loop pattern: iterate decisions, skip if comment_id is not an int or
  response is empty, call github.reply_to_review_comment, swallow per-comment
  exceptions. The only differences are the status == "addressed" skip and the
  dispositions filter, both one-liners. Consider consolidating into one shared
  function so GitHub reply behavior stays consistent.

src/forge/workflow/nodes/implement_review.py
eshulman2: Addressed in 8fe4b0d. The decision merge logic now lives in one shared merge_review_decisions helper used by both the implementation and proposal review paths, with unit coverage for replacement and legacy-item handling.

src/forge/prompts/v1/triage-proposal-review-threads.md
eshulman2: Addressed in 8fe4b0d. Proposal thread dispositions now emit forge_proposal_review_decisions_total labeled by artifact_type and disposition, along with a structured log containing the thread ID, disposition, and rationale. This gives us an early production signal for systematic classification drift.

src/forge/orchestrator/worker.py
eshulman2: Fixed in c27f1d2. Forge now stores the ID returned for its inline reply and correlates later human replies against either the original reviewer comment ID or Forge reply ID. Regression coverage models the reviewer replying directly to Forge.

src/forge/workflow/utils/proposal_review_threads.py
eshulman2: Fixed in c27f1d2. The proposal parser now excludes threads without an identity or comments before accessing the latest comment, with an empty-thread regression test.

src/forge/orchestrator/worker.py
eshulman2: Fixed in c27f1d2. Both PRD and spec paths now use the shared flatten_review_threads helper, which also safely skips empty threads.

src/forge/workflow/nodes/implement_review.py
eshulman2: Added coverage in c27f1d2 confirming a legacy review-objections file still posts the objection and pauses at review_response_gate.

src/forge/workflow/nodes/implement_review.py
eshulman2: Added coverage in c27f1d2 confirming previously accepted thread IDs are excluded from the next analysis while new threads remain.

src/forge/workflow/nodes/implement_review.py
eshulman2: Fixed in c27f1d2. Implementation and proposal replies now delegate to one shared reply_to_review_decisions helper.

src/forge/workflow/utils/proposal_review_threads.py
eshulman2: Fixed in c27f1d2 via the same shared reply helper; proposal-specific disposition and addressed filters remain configurable at the wrapper.

ekuris-redhat:
LGTM , I think it's safe to merge.

PR #239, opened 2026-07-30
feat(sandbox): surface agent transcript errors

### LINKED ISSUE #79: feat: surface agent transcript errors on container failure
## Problem

When a container exits non-zero, the worker logs container stderr — which for Claude Code is typically just MCP server startup noise, not the actual error. The real failure reason (tool refusal, context overflow, task misunderstanding) is buried in the JSONL agent transcript and never surfaced.

## Proposed solution

On non-zero container exit:

1. Locate the agent JSONL transcript in the workspace (e.g. from the Deep Agents runtime output)
2. Parse the last N assistant turns for error signals: refusals, `result: "error"` tool responses, `stop_reason: "max_tokens"`
3. Log the extracted error context at `logger.error` level alongside the exit code
4. Include the transcript path in the log so operators can inspect it directly

## Reference

Learned from fullsend: `internal/cli/run.go` lines ~839–847 — parses transcript JSONL on non-zero exit and emits `::error::` annotations to the workflow log.

iamdhrv:
## Change

- persist tool status and model response metadata in Forge's existing
  `.forge/history/<task>.json` artifact
- inspect the last three assistant turns plus intervening tool results after a
  non-zero container exit
- surface explicit tool errors, token-limit stops, and assistant refusals at
  error level
- always log the task-specific transcript path when history is available

## Why

Container stderr often contains only startup noise. Forge already persists the
Deep Agents conversation, but the host runner did not consult it and omitted
the stop/status metadata needed to identify failures. The new parser uses that
existing task-scoped JSON artifact, bounds logged excerpts to 1,000 characters,
and leaves startup unaffected when history is missing or malformed.

## Checks

- `uv run pytest -q tests/unit/sandbox/test_transcript_errors.py tests/unit/sandbox/test_runner_review_polling.py` (31 passed)
- `uv run ruff check src/forge/sandbox/runner.py containers/entrypoint.py tests/unit/sandbox/test_transcript_errors.py`
- Python byte-compilation for both changed runtime modules
- `git diff --check`

Fixes #79


eshulman2 (approved):
Rebased onto current main and reviewed the conflict resolution and resulting diff. The container-driver refactor is preserved, transcript extraction is wired to ExecResult, and focused tests plus Ruff checks pass. No blocking findings.

PR #243, opened 2026-08-02
Make sandbox runner pluggable with Kubernetes driver support

### LINKED ISSUE #30: Make sandbox runner pluggable with different execution drivers
## Problem

The current `ContainerRunner` (`src/forge/sandbox/runner.py`) is tightly coupled to Podman.
It calls `podman` directly, verifies Podman is in `PATH` at startup, and builds Podman-specific
CLI flags (`--rm`, `slirp4netns`, `-v` mounts with `:Z` labels, etc.). This limits Forge to
environments where Podman is installed and prevents deployment in production Kubernetes clusters
or other container runtimes.

## Proposed Solution

Extract an abstract `SandboxDriver` interface and refactor `ContainerRunner` to implement it
as the `PodmanDriver`. Additional drivers can then be added without touching the orchestrator.

### Interface sketch

```python
class SandboxDriver(Protocol):
    async def run(
        self,
        workspace_path: Path,
        task: TaskSpec,
        config: DriverConfig,
    ) -> ContainerResult: ...

    async def is_available(self) -> bool: ...
```

The orchestrator would accept a `SandboxDriver` and select the implementation via config
(e.g. `FORGE_SANDBOX_DRIVER=podman|k8s|docker`).

### Drivers to implement

- **PodmanDriver** (existing, refactored) — local rootless Podman, current default
- **KubernetesDriver** — run each task as a Kubernetes Pod; useful for production deployments
  where Forge itself runs in-cluster and workspaces are backed by PVCs or synced via init containers

### Kubernetes driver specifics

- Creates a `Job` (or `Pod`) in a configurable namespace
- Workspace synced to the pod via a PVC or `kubectl cp` / rsync init container
- Credentials passed as K8s `Secret` references, not plain env vars
- Respects resource `requests`/`limits` from driver config
- Collects logs via `kubectl logs` and maps pod exit codes to `ContainerResult`

## Acceptance Criteria

- [ ] `SandboxDriver` protocol/ABC defined with `run()` and `is_available()`
- [ ] `ContainerRunner` refactored into a `PodmanDriver` implementing the interface
- [ ] Driver selection via config (`FORGE_SANDBOX_DRIVER` env var or `settings.sandbox_driver`)
- [ ] `KubernetesDriver` implemented using the `kubernetes` Python client
- [ ] Existing unit tests pass; new tests cover driver selection and K8s driver logic
- [ ] `CLAUDE.md` / README updated to document driver configuration

eshulman2:
## Summary

- Extract `SandboxDriver` ABC from monolithic `ContainerRunner`, creating a pluggable driver interface (`ExecutionSpec` → `ExecutionResult`)
- Refactor Podman-specific logic into `PodmanDriver`
- Add `KubernetesDriver` that runs tasks as Kubernetes Jobs with PVC-based workspace sharing
- Select the driver with `FORGE_SANDBOX_DRIVER` (default: `podman`; also supports `kubernetes`)
- Add a Helm chart for the Forge API, worker, Redis, RBAC, OpenShift Route, and shared workspace PVC
- Support optional Google ADC Secret mounts in both the worker and sandbox Jobs
- Add driver-aware smoke testing and Kubernetes sandbox isolation

Closes #30

## Kubernetes deployment

The Helm chart under `charts/forge` provides API and worker Deployments, Redis Stack, namespace-scoped RBAC, a retained workspace PVC, an optional OpenShift Route, optional Google ADC propagation, and sandbox NetworkPolicies.

Sandbox Jobs:

- preserve the container entrypoint's exit code
- apply CPU and memory requests and limits
- disable service-account token mounting
- run with non-root, dropped-capability, RuntimeDefault seccomp security contexts
- default-deny ingress and egress, allowing DNS and public egress only for network-enabled executions

## Configuration

```bash
# Default
FORGE_SANDBOX_DRIVER=podman

# Kubernetes
FORGE_SANDBOX_DRIVER=kubernetes
WORKSPACE_BASE_DIR=/mnt/workspaces
K8S_NAMESPACE=forge
K8S_WORKSPACE_PVC=forge-workspaces
K8S_WORKSPACE_BASE_PATH=/mnt/workspaces
K8S_IMAGE_PULL_SECRETS=
K8S_SERVICE_ACCOUNT=forge-runner

# Optional Vertex AI credential Secret mounted into sandbox Jobs
K8S_GOOGLE_CREDENTIALS_SECRET=google-adc
K8S_GOOGLE_CREDENTIALS_KEY=forge-gcp-credentials.json
K8S_GOOGLE_CREDENTIALS_MOUNT_PATH=/etc/forge-gcp-credentials.json
```

## Validation

- [x] Ruff lint and formatting pass for changed Python files
- [x] 85 relevant sandbox and smoke tests pass locally
- [x] Helm lint covers default and Google-credentials-enabled variants
- [x] Helm resources deployed on a real OpenShift 4.20 cluster
- [x] API, worker, Redis, Route, RBAC, and RWO PVC deployed successfully
- [x] Worker selected `KubernetesDriver` using in-cluster authentication
- [x] Implementation sandbox Job completed successfully
- [x] Review sandbox Job completed successfully
- [x] The end-to-end workflow created an implementation PR
- [x] `forge smoke-test` now executes through the configured driver and creates Kubernetes workspaces on the shared PVC

The live OpenShift validation exposed and verified fixes for arbitrary-UID execution, Vertex ADC propagation, writable skills storage, Job status RBAC, resource quantity conversion, and workflow configuration inheritance.


src/forge/sandbox/drivers/kubernetes.py
```diff
@@ -0,0 +1,457 @@
...
+        with contextlib.suppress(AttributeError):
+            conditions = job.status.conditions or []
+            for cond in conditions:
+                if (
+                    cond.type == "Failed"
+                    and cond.status == "True"
+                    and cond.reason == "DeadlineExceeded"
+                ):
+                    return -1
+        return 1
```
eshulman2, line 440: The driver should preserve the entrypoint contract here. A failed Kubernetes Job does not expose its container exit code on Job status, so `_extract_exit_code()` currently collapses exit codes 2 (tests failed) and 3 (configuration error) to 1. Please read `pod.status.container_statuses[].state.terminated.exit_code` through the Core API before building `ExecutionResult`; otherwise `ContainerResult.tests_failed` and its error classification are wrong for this backend.

src/forge/sandbox/drivers/kubernetes.py
```diff
@@ -0,0 +1,457 @@
...
+            "--max-retries",
+            str(spec.max_retries),
+        ]
+        if spec.skip_tests:
+            container_args.append("--skip-tests")
+
+        memory_limit = _normalize_memory_quantity(spec.memory_limit)
+        resources = k8s_client.V1ResourceRequirements(
+            requests={"memory": memory_limit, "cpu": spec.cpu_limit},
+            limits={"memory": memory_limit},
```
eshulman2: `cpu_limit` is only applied as a request, while memory is applied as both request and limit. That changes the existing resource-limit contract and permits an agent Job configured for 2 CPUs to consume unbounded node CPU. Please include `"cpu": spec.cpu_limit` in `limits` and cover it in the manifest test.

src/forge/sandbox/drivers/kubernetes.py
```diff
@@ -0,0 +1,457 @@
...
+            name="forge-task",
+            image=spec.image,
+            args=container_args,
+            env=env_vars,
+            volume_mounts=volume_mounts,
+            resources=resources,
+            working_dir="/workspace",
+        )
+
+        pod_spec = k8s_client.V1PodSpec(
```
eshulman2, line 261: This generated sandbox pod needs an explicit isolation policy. At present `ExecutionSpec.network_mode` is ignored, the service-account token is automatically mounted, and the chart supplies no NetworkPolicy. Agent-controlled repository code can therefore reach Redis/API services in the namespace and the Kubernetes API. Please disable service-account token automount, apply an explicit pod/container security context, and add a NetworkPolicy that permits only the DNS/external traffic the sandbox requires.

src/forge/sandbox/drivers/kubernetes.py
```diff
@@ -0,0 +1,457 @@
...
+    async def execute(self, spec: ExecutionSpec) -> ExecutionResult:
+        from kubernetes import client as k8s_client
+        from kubernetes import config as k8s_config
+
+        k8s_config.load_config()
+        batch_api = k8s_client.BatchV1Api()
+        core_api = k8s_client.CoreV1Api()
+
+        job_name = self._job_name(spec.container_name)
+        prepared_spec, staging_dir = self._stage_external_mounts(spec)
```
eshulman2: This recursive staging copy runs synchronously on the asyncio worker thread. Large skill trees can stall unrelated worker activity. Also, staging happens before the `try/finally` in `execute()`, so if `_build_job_manifest()` raises, the copied directory is leaked. Please move preparation under the cleanup scope and offload potentially large copies to a worker thread (or make staging an explicit async operation).

eshulman2:
Additional merge-readiness issue outside the changed diff: `src/forge/smoke_test.py:127-136` still unconditionally requires Podman and checks for a locally present image. With `FORGE_SANDBOX_DRIVER=kubernetes`, the advertised smoke test therefore fails before exercising the Kubernetes driver. Please make Stage 2 driver-aware and add a Kubernetes execution smoke/e2e path; the PR description currently notes that a complete sandbox Job was not run against a real cluster.

I added a separate **Helm Lint** CI job in commit `ea58c8d`. It lints both the default chart and the Google-credentials-enabled variant.

eshulman2 (commented):
Re-review complete against `03f3001`. The previously raised implementation issues are addressed:

- failed Kubernetes Jobs now preserve the terminated container exit code;
- CPU is enforced as both a request and a limit;
- sandbox Jobs disable service-account-token automount, apply pod/container security contexts, and are selected by default-deny/external-egress NetworkPolicies;
- external mount staging is moved off the event loop and cleanup covers preparation failures;
- the smoke-test path is now driver-aware and places Kubernetes workspaces under the shared PVC base path.

Verification: both Helm variants lint and render successfully, focused sandbox/smoke tests pass (46 passed), and every PR CI check is green. I found no new merge-blocking code issue.

Residual rollout caveat: the Kubernetes smoke unit test mocks `ContainerRunner`, so it does not prove a complete Job against a live cluster. Given that the feature is opt-in and the implementation blockers are fixed, I consider this non-blocking for merge, but recommend completing the real-cluster Job validation before enabling the driver in production.

charts/forge/templates/api.yaml
```diff
@@ -0,0 +1,82 @@
...
+        {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      containers:
+        - name: api
+          image: {{ include "forge.image" . }}
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          securityContext:
```
eshulman2, line 32: The new vanilla-Kubernetes ownership setting is not applied to the API pod. The production Dockerfile uses the symbolic `USER forge`, while this container sets `runAsNonRoot: true`; without a numeric pod/container `runAsUser`, kubelet rejects it because it cannot verify that a named image user is non-root. `sandbox.runAsUser=1000` currently fixes only the worker and sandbox Job, so the README installation command still leaves the API unschedulable on vanilla Kubernetes. Please either apply the configured numeric UID to the API as well (preferably under an application-wide value), or make the image user numeric. The Kind gate should start/wait for the chart workloads rather than only server-dry-running them so this is exercised.

eshulman2 (commented):
Fresh review against `4b28c2e` found one merge blocker in the newly added vanilla-Kubernetes path: `sandbox.runAsUser` is applied to worker/sandbox pods but not the API, even though the application image declares the symbolic user `forge` and the API sets `runAsNonRoot`. A vanilla kubelet will reject that API container because no numeric non-root UID is supplied. The Kind gate only performs server-side dry-run for chart resources and therefore does not detect workload startup failures. I left the exact fix request inline.

Everything else reviewed cleanly: the real Job/PVC/exit-code Kind tests are valuable, all eight CI checks are green, both Helm configurations lint/render successfully, and the focused local suite had 90 passes plus 3 known unrelated `review` import failures.

eshulman2:
Addressed in `776379a`. Runtime UID/GID configuration is now application-wide via `runtimeSecurity`, so the API, worker, and dynamically created sandbox Jobs receive the same numeric `runAsUser`/`fsGroup`. The Kind gate now builds the application image, installs the Helm chart with `--wait`, verifies both API and worker deployments become ready and run as UID 1000, then runs the Kubernetes driver tests. All eight checks, including the enhanced Kind gate, are green.

eshulman2 (commented):
Recheck complete against `776379a`. The API UID blocker is resolved: `runtimeSecurity` now applies the numeric UID/fsGroup consistently to API and worker pods and passes the same values to sandbox Jobs. The Kind gate now builds the real application image, installs the Helm release with `--wait`, verifies API/worker rollout, and asserts the API runs as UID 1000, so it covers the failure mode raised in the prior review.

Local verification also passes: both Helm variants lint, the rendered API pod has `runAsUser: 1000` and `fsGroup: 1000`, focused tests pass (48 passed), and the full PR diff passes whitespace checks. All eight required CI checks are green. No remaining merge-blocking findings; this is merge-ready.

PR #280, opened 2026-08-07
Make container command timeout configurable

eshulman2:
## Summary
- add a positive `CONTAINER_COMMAND_TIMEOUT` setting with a 600-second default
- pass it into task containers and apply it to implementation commands, reviewer commands, and fallback tests
- show the effective timeout in the container system prompt
- document the setting and validate invalid values safely

## Testing
- Ruff passed for all changed Python files
- 71 focused tests passed


eshulman2:
### Code review

Found 1 issue:

1. Reviewer command timeout silently changes from 300s to 600s, and the per-call defaults become dead code. `command_timeout(default)` only returns `default` when `CONTAINER_COMMAND_TIMEOUT` is unset, but `runner.py` always sets that env var from `container_command_timeout` (default 600). So `command_timeout(300)` for the reviewer never returns 300 in a real container run — it returns the single configured value (600 by default), doubling the reviewer's previous 5-minute timeout and making the differing defaults (600 vs 300) meaningless outside tests. If the intent was to keep the reviewer at 300s, add a separate setting or pass an explicit value; otherwise the 300 default is misleading.

https://github.com/forge-sdlc/forge/blob/f702fec2869c522bb44d48e235737ddb9c4c1ae4/containers/entrypoint.py#L738-L741

The env var is always populated here, so the fallback default is never used in production:

https://github.com/forge-sdlc/forge/blob/f702fec2869c522bb44d48e235737ddb9c4c1ae4/src/forge/sandbox/runner.py#L212-L214

eshulman2:
Addressed the review feedback in 0483837:

1. **Reviewer timeout** — `run_reviewer_agent` now uses `command_timeout(600)`, the single configured `CONTAINER_COMMAND_TIMEOUT`. The previous `command_timeout(300)` fallback was dead code (`runner.py` always sets the env var), so the reviewer already ran at 600s by default; this makes the intent explicit and uniform across agent tasks, reviewer, and fallback tests.
2. **Test-timeout log** — compute the timeout once and reuse it for both `subprocess.run` and the timeout log, removing the redundant env re-read.
3. **Validation** — added a `validate_container_timeouts` model validator that rejects `container_command_timeout > container_timeout` at startup.

Docs (`.env.example`, `containers/README.md`) now state the timeout covers agent tasks, reviewer commands, and fallback tests and must not exceed `CONTAINER_TIMEOUT`. Added `tests/unit/test_config_container.py` for the new validation.

PR #302, opened 2026-08-11
[AISOS-2392] Forge ignores PR review comments when developer shares the bot's GitHub account  

### LINKED ISSUE #246: dev: Forge ignores PR review comments when developer shares the bot's GitHub account
## Problem

When developing locally with a personal GitHub account used as the Forge bot token, submitting a PR review comment as yourself is silently dropped by Forge. No error is surfaced — the workflow simply does not advance.

**Root cause:** Forge detects its own comments by comparing `sender.login` from the webhook payload against the login returned by `GET /user` (the authenticated GitHub token owner). In a dev setup where the developer's GitHub account *is* the bot account, these are always equal — so every comment the developer posts looks like a Forge self-comment and is skipped.

This affects three filter sites in the worker:
- Inline PR review comments (`review_response_gate`)
- Issue comments on PRD proposal PRs
- Issue comments on spec proposal PRs

## Proposed solutions

Three options for team discussion — they are not mutually exclusive.

### Option A — Comment prefix (`FORGE_COMMENT_PREFIX`)

Forge signs every GitHub comment it posts with a configurable hidden prefix (e.g. `<!-- forge-bot -->`). Self-comment detection checks for the prefix in the comment body rather than comparing logins.

- New config: `FORGE_COMMENT_PREFIX` (empty = disabled, falls back to login check)
- On post: all `create_issue_comment` and `create_review_comment` calls prepend the prefix
- On receive: comment body starts with prefix → skip; otherwise → process
- The `<!-- ... -->` HTML comment syntax renders invisible in GitHub markdown

This is the most targeted fix. A developer's hand-written comments never carry the prefix, so they're never dropped, regardless of which GitHub account Forge uses.

### Option B — Explicit bot login override (`FORGE_GITHUB_BOT_LOGIN`)

New config `FORGE_GITHUB_BOT_LOGIN` hard-codes the GitHub username Forge compares against instead of calling `GET /user` at runtime.

- In production: set to the dedicated bot account login
- In dev: leave empty to disable the login check entirely (all comments are processed)
- Pro: no changes to comment content
- Con: disabling the check means Forge has no self-comment protection in dev — potential for Forge to react to its own comments if the config is accidentally left unset in production

### Option C — Dedicated GitHub App

Switch from a PAT to a GitHub App for local development as well. Comments posted by a GitHub App carry a `[bot]` suffix in the sender login (e.g. `forge-dev[bot]`), giving Forge a distinct identity from the human developer. The existing login comparison already handles this case correctly.

- Pro: zero code changes; proper identity separation at the platform level
- Con: heavier one-time setup (register App, install on repos, configure private key + app ID); less practical for a fast local dev loop

## Suggested default

Option A as the code-level fix, with Option C as the longer-term architecture for teams that want a clean prod/dev identity boundary.

ekuris-redhat:
## Summary

This pull request implements an automated outbound comment signing mechanism and a dual-check webhook self-comment filtering strategy for the Forge bot. By appending or prepending a configurable HTML comment signature to outbound GitHub posts, the system can reliably distinguish its own automated updates from human feedback, preventing infinite webhook loops even when sharing identical credentials or personal access tokens (PATs) in local development environments.

### Configuration & Settings
- **`src/forge/config.py`**: Added the `forge_bot_comment_prefix` field of type `str` (defaulting to `""`) to the `Settings` class with Pydantic `Field` validation.
- **`.env.example`**: Exposed and documented the `FORGE_BOT_COMMENT_PREFIX` environment variable under the GitHub configuration section.
- **`docs/reference/config.md`**: Updated documentation to detail the purpose and utilization of the new signature prefix, clarifying that it is intended to enable development and testing with the same API keys without webhook loops, and should not be used in production.

### Core Bot Signature Utilities
- **`src/forge/workflow/utils/automated_review_triage.py`**:
  - Implemented `prepend_bot_prefix` to safely prepend the bot signature (wrapped inside HTML comments `<!-- ... -->` if not already wrapped) and guard against double-prepending and `None` body payloads.
  - Implemented `is_self_comment` with $O(1)$ complexity. It applies a dual-check rule when a prefix is configured (matching username *and* body must start with prefix) and falls back to username equality if the prefix is empty. It also ensures GitHub App accounts ending in `[bot]` are correctly validated against the Forge bot login base, preventing non-Forge bots from being mistakenly ignored as self-comments.

### Outbound Client Integration
- **`src/forge/integrations/github/client.py`**: Modified low-level comment and reply methods (`create_review_comment`, `create_issue_comment`, and `reply_to_review_comment`) to dynamically sign outgoing comments using the configured signature prefix.

### Orchestrator Worker Event Handlers
- **`src/forge/orchestrator/worker.py`**: Integrated `is_self_comment` filtering across all event loop endpoints (inline review comment replies, PRD issue comments, spec issue comments, PR review events, and proposal replies) to replace fragile, direct username equality comparisons. Enabled safe loading of settings using a `getattr` fallback.

### Testing Suite
- Created `tests/unit/test_config_bot_signature.py` to verify configuration defaults and loading dynamics.
- Created/updated unit tests in `tests/unit/workflow/utils/test_automated_review_triage.py` to assert edge-case scenarios (SC-001 through SC-004) for prefix checks, fallback username matching, and bot suffix rules.
- Created `tests/unit/integrations/github/test_outbound_signature.py` to verify client-level comment signing behavior.
- Updated integration and webhook handlers in `tests/unit/orchestrator/test_worker.py` and `tests/unit/api/routes/test_github_webhook.py` to verify loop prevention rules, and configured `tests/conftest.py` with dummy environment variables to prevent settings validation errors during pytest startup.
- Created `tests/unit/orchestrator/test_worker_prd_pr.py` to cover the new signature-checking worker handlers.

## Implementation Notes

- **Robust Null-Safety**: Both `prepend_bot_prefix` and `is_self_comment` coerce falsy or `None` payloads to empty strings to avoid runtime `AttributeError` exceptions when parsing raw webhook events.
- **Whitespace & Spacer Resilience**: The signature lookup parser checks for both space-omitted (`<!--prefix-->`) and standard spaced (`<!-- prefix -->`) comment structures to guarantee compatibility regardless of automated formatting.
- **Developer Local Loop Support**: When a signature prefix is configured, a comment coming from the bot's username that does *not* contain the prefix is successfully processed as a human comment. This allows developers to use a single bot PAT for both automated workflows and local manual testing without being blocked by loop prevention. Note that this configuration is intended strictly for development and testing environments and should not be used in production.

## Testing

- **Unit Testing**: Executed extensive unit tests covering the configuration settings, the prepend wrapper functionality, and the core `is_self_comment` evaluation logic.
- **Integration Testing**: Verified that orchestrator worker events successfully parse webhook deliveries, filter out authentic self-comments, and proceed with human developer actions.
- **Linting & Code Quality**: Validated that all new/modified files conform cleanly to `mypy` and `ruff` formatting guidelines.

## Related Tickets

- [AISOS-2392](https://redhat.atlassian.net/browse/AISOS-2392)
- [AISOS-2395](https://redhat.atlassian.net/browse/AISOS-2395)
- [AISOS-2396](https://redhat.atlassian.net/browse/AISOS-2396)
- [AISOS-2397](https://redhat.atlassian.net/browse/AISOS-2397)
- [AISOS-2398](https://redhat.atlassian.net/browse/AISOS-2398)
- [AISOS-2399](https://redhat.atlassian.net/browse/AISOS-2399)
- [AISOS-2400](https://redhat.atlassian.net/browse/AISOS-2400)
- [AISOS-2401](https://redhat.atlassian.net/browse/AISOS-2401)

---
*Generated by [Forge](https://github.com/forge-sdlc/forge) SDLC Orchestrator*

## Auto-Review Notes

The following review criteria could not be resolved after all retry attempts.
Human reviewers should pay particular attention to these areas.

### implement_task — AISOS-2399
**Skill:** implement-task | **Retries:** 2/2 exhausted

> Verdict could not be parsed

Fixed #246


ekuris-redhat:
### 🛠️ Forge PR Commands

This pull request was created by Forge! You can use the following commands by commenting on this PR:

*   `/forge rebase` - Merge the base branch (e.g. `main`) into this PR branch, with conflicts resolved by AI.
*   `/forge skip-gate <name>` - Skip a named CI check (substring match) for this PR. This setting persists across subsequent pushes.
*   `/forge unskip-gate <name>` - Remove a previously set CI check skip.

Feel free to use these commands to manage your workflow!

docs/reference/config.md
```diff
@@ -19,6 +19,7 @@ All configuration is via environment variables in `.env`. See `.env.example` in
 |----------|-------------|
 | `GITHUB_TOKEN` | Personal Access Token with `repo` and `read:org` scopes |
 | `GITHUB_WEBHOOK_SECRET` | Secret for validating GitHub webhook signatures |
+| `FORGE_BOT_COMMENT_PREFIX` | Prefix to add to all comments made by the Forge bot (e.g., signature or identifier), also used for webhook self-comment filtering to prevent loops |
```
eshulman2: please explain that this is meant to allow development with the same user api keys that are used to comment and shouldn't be used in production.

eshulman2 (changes requested):

ekuris-redhat:
I took some time to put this change through a proper e2e run before signing off, so here is a rundown of what I actually checked.

The situation this PR fixes is the one where a developer runs Forge with the same GitHub account the bot posts from, and Forge ends up ignoring that person's review comments because it assumes they came from itself. I wanted to confirm both of the fix really behave: the outbound signing and the inbound filtering.

First I ran the test suite locally and everything came back green, the full unit set passed with no failures.

Then I went for a live check against this very PR. Using the Forge GitHub client I posted a comment where the caller passed no signature at all. When I pulled the stored comment back from GitHub, the body had already been wrapped with the hidden signature marker on its own, exactly as intended. It lands as an HTML comment, so it stays invisible in the rendered view but is clearly there in the raw body. I removed that test comment afterwards so it would not clutter the thread.

Next I exercised the inbound side using my own account, which also happens to be the account the bot posts from, so it reproduces the shared account case precisely. Three scenarios, all behaved the way they should:

1. A comment that Forge itself wrote, carrying the signature, was correctly recognized as its own and skipped.
2. A comment I typed by hand from that same account, with no signature, was correctly treated as real human feedback and picked up. This is the exact case that used to break.
3. With the signature switched off, the old behavior comes back and the comment gets swallowed, which lines up perfectly with the bug this PR closes.

One small thing I bumped into while testing locally: the suite reads a populated .env file, so on a developer machine that already has the signature or model settings configured, a couple of the settings tests can trip. It runs clean in CI since there is no .env there. Nothing to do with the feature itself, just a test isolation nuance worth flagging.

All in all this looks solid to me and matches the intended behavior. Nice work on this one.
Worker Log: https://docs.google.com/document/d/1B2K5A5yePI7hR6t-I2AlcmyB_Dfs-mrELo_DDgfabC8/edit?tab=t.0

eshulman2:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

PR #303, opened 2026-08-17
security: isolate host agents and sanitize subprocesses

eshulman2:
## Summary

- isolate host agents in a dedicated virtual, read-only filesystem with no shell backend
- enforce safe built-in tools and exact default-deny MCP tool allowlisting
- stage validated skills into the agent root and reject traversal and symlinks
- prevent implementation and reviewer shells from ordinarily inheriting container secrets
- pin Deep Agents 0.6.12 and document production-safe Podman, service-container, and future Kubernetes deployment controls

## Security boundary

This phase intentionally leaves the top-level container environment, provider authentication, Google ADC mounts, Langfuse callbacks, tracing metadata, session correlation, and dynamic-skill behavior unchanged. `inherit_env=False` is defense in depth: sufficiently capable code may still inspect container-level secrets through mechanisms such as `/proc`.

## Validation

- `uv run ruff check src/forge/integrations/agents src/forge/skills/resolver.py src/forge/main.py src/forge/orchestrator/worker.py src/forge/config.py src/forge/cli.py containers/entrypoint.py tests/unit/integrations/agents`
- targeted agent, skill, container entrypoint, sandbox backend, worker, and configuration unit suites
- `git diff --check origin/main...HEAD`

## Deployment docs

See `docs/security/agent-isolation.md` for configuration, rootless Podman controls, hardened service-container settings, Kubernetes prerequisites, negative tests, canary rollout, incident response, and deferred gateway/proxy work.


eshulman2:
### Code review

Found 1 issue:

1. `validate_agent_root()` calls `root.mkdir(parents=True, exist_ok=True, mode=0o700)`, but `mkdir` ignores `mode` when the directory already exists (and the process umask can weaken it on creation). If the agent root pre-exists with permissive permissions (e.g. `0o755`/`0o777`), validation still passes and the isolation boundary this PR establishes is silently weakened — other local users/processes could read the agent workspace and skill tree or plant skills. There is no follow-up `chmod`. Consider an explicit `root.chmod(0o700)` after `mkdir`. (bug — undermines the PR's own isolation goal)

https://github.com/forge-sdlc/forge/blob/5faf4ed748ba96edf8337d805abe5ae9b8f3161b/src/forge/integrations/agents/security.py#L49-L55

eshulman2:
Addressed the existing review feedback in f71b3d2:

- enforce `0700` on an existing or newly created agent root, after overlap validation so rejected paths are never chmodded
- add regression coverage for permission enforcement and rejected-path behavior

Independent review also found and fixed an MCP/built-in name collision: an allowlisted MCP tool named `execute`, `write_file`, or `edit_file` could cause the additive Deep Agents built-in with the same name to pass the name-based middleware. Those identifiers now fail closed, with regression coverage and documentation.

Validation: targeted Ruff checks pass; all 2,149 unit tests pass (one pre-existing AsyncMock warning). No additional blocking findings in the PR diff.

src/forge/integrations/agents/agent.py
```diff
@@ -412,12 +389,30 @@ async def _load_mcp_tools(self) -> list[Any]:
...
+            if item.strip()
+        }
+        collisions = {
+            identifier
+            for identifier in allowed
+            if identifier.rpartition(":")[2] in PROHIBITED_BUILTIN_TOOLS
+        }
+        if collisions:
+            names = ", ".join(sorted(collisions))
+            raise ValueError(f"MCP tool names collide with prohibited host tools: {names}")
```
eshulman2, line 409: **Bug: MCP tool names are only checked against `PROHIBITED_BUILTIN_TOOLS`, not `SAFE_BUILTIN_TOOLS` — an allowlisted MCP tool can silently shadow a trusted local tool.**

This collision check rejects an `AGENT_MCP_ALLOWED_TOOLS` entry whose bare tool name is `write_file`/`edit_file`/`execute`, but does nothing if the name is `ls`/`read_file`/`glob`/`grep` (`SAFE_BUILTIN_TOOLS` in `security.py:14`).

I traced how `tools=` and middleware tools are merged in the installed `deepagents`/`langchain`/`langgraph` versions:
- `langchain/agents/factory.py:876,927`: `available_tools = middleware_tools + regular_tools` — `middleware_tools` come from `FilesystemMiddleware` (the real sandboxed `ls`/`read_file`/`glob`/`grep`), `regular_tools` is `tools=mcp_tools` passed below at line 491, which comes *after*.
- `langgraph/prebuilt/tool_node.py:781`: `self._tools_by_name[tool_.name] = tool_` — a plain dict keyed by name, so the later entry (the MCP tool) silently overwrites the earlier one (the sandboxed filesystem tool) when both share a name.

Concretely: an operator sets `AGENT_MCP_ALLOWED_TOOLS=someserver:read_file` (plausible — many MCP servers expose generically named tools). `parse_host_tools`/this collision check both pass with no error. At runtime, the model's `read_file` calls are then serviced by the external MCP implementation instead of the isolated filesystem backend, silently defeating the virtual-root sandbox this PR builds.

Suggest extending this check (and the one in `HostToolAllowlistMiddleware`'s allowed-set construction at line 493) to also reject/flag MCP identifiers whose bare name is in `SAFE_BUILTIN_TOOLS`, not just `PROHIBITED_BUILTIN_TOOLS`.

src/forge/integrations/agents/security.py
```diff
@@ -0,0 +1,144 @@
...
+    if prohibited:
+        raise ValueError(f"Prohibited host agent tools: {', '.join(sorted(prohibited))}")
+    return requested
+
+
+def validate_agent_root(root: Path, project_root: Path, workspace_base: str = "") -> Path:
+    """Create and validate a dedicated agent root that cannot expose Forge data."""
+    if root.is_symlink():
+        raise ValueError(f"Agent root must not be a symlink: {root}")
+    root.mkdir(parents=True, exist_ok=True, mode=0o700)
```
eshulman2, line 54: **Minor: TOCTOU between the symlink check and `mkdir`.**

`root.is_symlink()` is checked, then `root.mkdir(parents=True, exist_ok=True, mode=0o700)` runs as a separate step. `Path.mkdir(exist_ok=True)` treats an existing symlink-to-directory as already existing (it follows the symlink for the existence/`is_dir()` check), so a local actor with write access to the parent directory could swap this path for a symlink in between the two calls. The later `resolve()` + protected-path checks (lines 70-74) still catch the well-known sensitive targets (`.env`, `.git`, `.ssh`, etc.), so this doesn't fully bypass the boundary, but an attacker-controlled target outside that explicit denylist would pass silently. Low severity — exploiting it requires an actor already at roughly the same filesystem trust level as the Forge process — but worth a `O_NOFOLLOW`-style atomic check if you want to close it fully. Not a blocker.

src/forge/integrations/agents/security.py
```diff
@@ -0,0 +1,144 @@
...
+            raise ValueError(f"Skill path escapes its source: {entry}") from exc
+    return resolved_source
+
+
+def initialize_agent_skills(agent_root: Path, source_root: Path) -> Path:
+    """Seed the isolated runtime skill tree from committed skill directories."""
+    safe_source = _assert_safe_tree(source_root)
+    skills_root = agent_root / "skills"
+    skills_root.mkdir(parents=True, exist_ok=True)
+    for source in sorted(safe_source.iterdir()):
```
eshulman2, line 120: **Note: stale skills are never pruned.**

`shutil.copytree(..., dirs_exist_ok=True)` only adds/overwrites; it never removes a skill directory that existed in a previous run but was deleted from the committed source. On a long-lived `AGENT_ROOT_DIR` (the documented production pattern, e.g. `/var/lib/forge/agent` outside Kubernetes' ephemeral `emptyDir`), a skill intentionally removed upstream (e.g. found unsafe) stays available to the host agent until an operator manually clears `agent_root_dir/skills`. This looks covered by the "Deferred security work" section of `docs/security/agent-isolation.md` (tightening dynamic-skill trust/pinning is called out as future work), so treating this as a known, disclosed limitation rather than a blocker — just flagging it here since it's easy to miss.

eshulman2 (commented):
Automated inline review. Overall this is a solid hardening pass (isolated host-agent root, exact tool/MCP allowlisting, sanitized subprocess env, symlink/traversal guards) and CI is green. One finding below is verified against the actual `langchain`/`langgraph` tool-merge code and should be fixed before merge since it undermines this PR's own stated goal (exact allowlisting instead of trusting names). The other two are lower-severity notes, not blockers.

eshulman2 (commented):
Re-reviewed after the latest push. Commit ad58decb correctly fixes both prior findings: the MCP/safe-builtin-tool collision check (`agent.py:404`, now checks `SAFE_BUILTIN_TOOLS | PROHIBITED_BUILTIN_TOOLS`, with a new test) and the stale-skill pruning (`security.py`, committed skills now rebuilt exactly into a separate `committed-skills` dir each startup, with tests). Good fixes.

However, this same commit introduces a new regression by only updating the *host-agent* skill path and missing the *container-agent* one — see inline comment. Also note CI's Lint job is currently red on this commit (ruff format check on `agent.py`), which is a mechanical fix, not a design issue.

src/forge/sandbox/runner.py
```diff
@@ -270,24 +270,17 @@ def _get_skill_mounts(
...
 
         Returns:
             Tuple of (mounts, container_paths) where:
             - mounts: List of (host_path, container_path) tuples
             - container_paths: Comma-separated paths for AGENT_SKILL_PATHS env var
         """
-        skills_dir = Path.cwd() / self.settings.skills_dir.rstrip("/")
-        # Assumes worker and runner share the same host filesystem — skills_install_dir
-        # is populated by the worker and mounted into the container from the host.
+        skills_dir = self.settings.skills_install_dir
```
eshulman2: **Regression: container tasks will silently stop getting default/committed skills mounted.**

`initialize_agent_skills` (this commit) now writes committed default/project skills into `AGENT_ROOT_DIR/committed-skills`, and leaves `AGENT_ROOT_DIR/skills` for runtime-fetched packages only (see `agent.py`'s updated `_get_skill_paths`, which was correctly changed to pass both dirs: `resolve_skill_paths(ticket_key, root_dir / "committed-skills", skills_install_dir=root_dir / "skills")`).

This call site wasn't updated to match: `skills_dir = self.settings.skills_install_dir` (`= agent_root/skills`) is passed as `resolve_skill_paths`'s first positional arg with no `skills_install_dir=` kwarg. `resolve_skill_paths` treats that first arg as the base containing `default/` and `{project}/` — but those now live under `committed-skills`, not `skills`. Concretely: `default_dir = agent_root/skills/default` will never exist post-fix, so `host_path.exists()` is `False`, the mount is skipped with a `logger.warning`, and every container implementation/reviewer task silently loses its default skill set (and any committed per-project override). Only skills that happen to be runtime-fetched into `agent_root/skills/{project}` would still resolve (mislabeled as "override" but functionally the fetched-skills branch).

No test currently exercises `_get_skill_mounts` against real directories, so CI won't catch this. Suggest mirroring `agent.py`'s fix here: `resolve_skill_paths(ticket_key or "", agent_root / "committed-skills", skills_install_dir=agent_root / "skills")` (may need a new `settings` property for the committed-skills path, analogous to `skills_install_dir`, since `self.settings.skills_install_dir` currently only exposes the runtime-fetched location).

eshulman2: Fixed in f923ef5. Container skill resolution now uses the committed snapshot as the base and the runtime install directory as the fetched layer. Added a filesystem-backed regression test covering default, committed project, and fetched project mounts.

src/forge/integrations/agents/agent.py
eshulman2: Fixed in ad58decb: collisions now include both safe and prohibited built-in names, with regression coverage.

src/forge/integrations/agents/security.py
eshulman2: Fixed in ad58decb: committed skills are rebuilt into a separate snapshot directory on startup, pruning stale committed entries while preserving runtime-fetched skills.

src/forge/integrations/agents/security.py
eshulman2: Acknowledged as a non-blocking local-filesystem TOCTOU hardening opportunity. The current deployment boundary requires the agent-root parent to be controlled by the Forge service account; atomic no-follow directory creation remains follow-up hardening.

Issue #162, opened 2026-07-16
Support configurable source providers and mixed-provider repository workflows


eshulman2:
## Problem

Forge currently models repositories primarily as GitHub-shaped `owner/repo` values and keeps workflow-level fields such as `current_repo`, `current_pr_number`, and `pr_urls`. This cannot safely represent a workflow spanning repositories on different source providers or provider instances—for example, a GitHub PR and a self-managed GitLab MR that happen to share native number `42`.

The product roadmap in PR #157 defines the intended direction: the source provider and connection belong to each repository, not to the workflow globally, and each repository change progresses independently under aggregate workflow gates.

## Proposed capability

Introduce configurable source-control connections and canonical repository references, then place the existing GitHub behavior behind a provider-neutral contract.

Example repository configuration:

```yaml
id: payments-api
provider: github
connection: public-github
namespace: acme/payments
default_branch: main
change_request_mode: fork
```

A connection should hold provider-specific API/base URLs, credential references, webhook verification configuration, TLS/CA configuration, proxy settings where applicable, and allowed namespaces. Credentials must not be embedded in Jira metadata or agent context.

Workflow state should track a map of per-repository work items rather than one global current PR:

- canonical repository identity and configured connection
- source revision and working branch
- PR/MR identity, URL, and state
- provider-normalized CI/check state
- review state
- execution, retry, and error state
- explicit dependencies on other repository work items

The durable identity of a change request should be composite, such as `(connection, repository_id, native_id)`, because native PR/MR numbers are provider-local.

## Delivery slices

1. Define `SourceControlProvider`, repository reference, connection, normalized event, change-request, review, and check-run contracts.
2. Adapt existing GitHub behavior to those contracts without regression.
3. Migrate workflow checkpoints from single-current-PR fields to per-repository work items.
4. Add GitLab.com support.
5. Add configurable self-managed GitLab connections, including private CA and capability discovery.
6. Support one workflow spanning mixed GitHub/GitLab repositories with independent change requests and aggregate gates.

## Acceptance criteria

- Repository configuration selects a provider and named connection independently for each repository.
- No workflow node imports a concrete GitHub or GitLab client.
- GitHub behavior passes the provider contract suite without regression.
- Interleaved/replayed events update only the addressed repository work item.
- PR/MR routing uses composite identity rather than native number or title parsing.
- Aggregate CI, review, and completion policy is derived explicitly across required repository changes.
[...]

PR #311, opened 2026-08-19
Add Git Platform Provider Contracts

danchild:
## Summary

Implements interface contracts to support multiple git platform providers (design in #162): a provider-neutral source control abstraction layer that enables Forge to work with GitLab, GitHub, and future platforms without coupling workflow logic to any single provider's API.

- Add a source control exception hierarchy (`SourceControlError` and subclasses) so adapters can translate provider-specific failures into a common vocabulary
- Add normalized data models (enums and dataclasses) defining the provider-neutral shapes for repositories, change requests, reviews, checks, and webhook events
- Add the `SourceControlProvider` protocol defining the contract GitHub and GitLab adapters must implement
- Add `repos.yaml` configuration loading and validation to the registry, allowing deployments to register connections to GitLab or additional GitHub instances
- Add `Registry.resolve()` for turning an identifier into a usable adapter, with fallback to an implicit `github-default` connection so existing GitHub-only deployments keep working without touching `repos.yaml`
- Add `register_adapter_factory()` so provider packages can register their adapter constructors without the registry importing provider-specific code

This segment establishes the contracts and configuration machinery; segment 2 will cut the existing GitHub integration over to the registry, and segment 3 will add the GitLab adapter itself.

## Implementation Notes

- **Zero-config GitHub fallback**: An implicit `github-default` connection (via `GITHUB_TOKEN` / `GITHUB_WEBHOOK_SECRET`) ensures unregistered GitHub namespaces still resolve, so the registry can be adopted incrementally without breaking existing deployments
- **Fail-fast validation**: Config parsing includes comprehensive up-front validation (provider names, required fields, `allowed_namespaces` membership, provider/connection/namespace uniqueness) so a bad `repos.yaml` fails at load time with a clear `ProviderConfigError` instead of surfacing as a confusing runtime failure
- **Adapter factory registration**: The registry accepts adapter factories via `register_adapter_factory()` rather than importing provider code directly, keeping the abstraction layer clean and avoiding circular dependencies

## Testing

- Comprehensive test coverage for the exception hierarchy, data models, protocol conformance, config loading/validation, and resolution logic
- Each rejection path (unknown provider, missing credentials, namespace collisions, malformed YAML) has explicit test coverage
- Registry tests cover resolving by explicit repository ID, namespace with provider hint, and implicit default connection


src/forge/integrations/source_control/registry.py
```diff
@@ -0,0 +1,276 @@
...
+    """Register the adapter constructor a provider's connections resolve to.
+
+    Called by each provider's integration package at import time (the GitHub
+    adapter registers itself in segment 2) — contracts.py and registry.py stay
+    free of provider-specific imports.
+    """
+    _ADAPTER_FACTORIES[provider] = factory
+
+
+def _build_implicit_connections(_settings: Settings) -> dict[Provider, Connection]:
```
eshulman2: Could the implicit GitHub path use `settings.github_token` (or otherwise honor the configured settings source)? `Settings` can load `GITHUB_TOKEN` from `.env` without adding it to `os.environ`, so the direct environment check in `resolve()` can reject an otherwise valid existing configuration.

src/forge/integrations/source_control/registry.py
```diff
@@ -0,0 +1,276 @@
...
+            )
+
+        connections[name] = Connection(
+            name=name,
+            provider=provider,
+            base_url=entry.get("base_url", ""),
+            credential_env=credential_env,
+            webhook_secret_env=entry.get("webhook_secret_env", ""),
+            ca_path=entry.get("ca_path"),
+            allowed_namespaces=entry.get("allowed_namespaces"),
```
eshulman2: Could we validate that `allowed_namespaces` is a `list[str]` here? A YAML string is currently accepted, and the membership check below then uses substring semantics—for example, `acme/pay` matches `acme/payments`.

danchild: valid

PR #312, opened 2026-08-19
Implement the GitHub source control adapter

danchild:
## Summary

This implements GitHub as the first concrete `SourceControlProvider`, against the conformance contract added in #311, and cuts the webhook ingestion path, the repo-resolution path, the queue transport, and the worker's event-handling logic over to it. Before this, GitHub-specific logic was scattered across raw webhook payload parsing, hardcoded `github.com` URLs, and `EventSource.GITHUB` string checks; afterward, everything flows through a single typed `NormalizedEvent` and a registry-resolved adapter, from webhook receipt through to git operations and workflow routing.

## What changed

**GitHub adapter (`GitHubAdapter`)**
- Webhook verification (HMAC-SHA256, failing closed when no secret is configured) and parsing into `NormalizedEvent`, including a normalized `check_suite_status` field so callers don't read raw check-suite JSON
- Repository/identity operations, change request CRUD, comments and replies, paginated review submission retrieval, check-run/check-artifact retrieval, and file/branch operations — filling in the full `SourceControlProvider` contract
- Hardened along the way: missing PR numbers and null comment authors are rejected instead of silently defaulting, oversized files and stale-sha writes are surfaced as clear errors instead of corrupting state or looping forever, check-run log fetching is gated on the GitHub Actions app specifically so other Apps using the Checks API (CodeQL, Codecov, etc.) aren't misdirected, and provider-error translation now covers network-level transport failures and non-numeric `Retry-After` values in addition to HTTP status codes

**Source control registry and resolver**
- `registry.py` loads the optional `repos.yaml`-shaped config at process startup (failing app startup on a misconfigured file rather than the first webhook) and resolves a repository identifier to a connection and its adapter, falling back to an implicit zero-config GitHub connection when a namespace isn't registered
- `workflow/utils/source_control.py` gives workflow nodes a single `resolve_repository` / `get_adapter` entry point so nodes never import a concrete client or re-derive the resolve pattern themselves

**Webhook route and queue**
- The webhook route now goes through the adapter resolved for the incoming repository and `QueueProducer.publish_event` instead of the old ad hoc parsing path, preserving delivery dedup and the event-id fallback
- `EventSource.GITHUB` is renamed to `SOURCE_CONTROL` throughout, and a `NormalizedEvent` queue transport (now including `CheckRun.output`) carries the full typed event across the Redis hop

**Workflow nodes**
- Comment-posting nodes (Q&A answers, model policy errors, human review notices, PRD/spec proposal PRs) resolve the repo through the registry and go through the adapter instead of constructing and closing a `GitHubClient` directly
- `GitOperations` takes connection-scoped credentials instead of a process-wide token and a hardcoded `github.com` host: clone/fork/remote URLs are built from the connection, `GIT_SSL_CAINFO` is set when a CA bundle is configured (so Enterprise Server with a self-signed cert works), and `prepare_workspace` is now async so it can resolve the write target through the adapter
- A guard test (`test_no_concrete_client_imports.py`) walks the workflow node modules and fails if any of them import the concrete `GitHubClient`, so the adapter boundary can't quietly erode in a future change

**Worker**
- Every raw-payload/`event_type`-substring detection block in `_handle_resume_event` reads typed `NormalizedEvent` fields instead — CI-webhook detection, `/forge` skip-gate/unskip-gate/rebase commands, PRD-PR and spec-PR review/merge/comment handling, inline review replies, human-review-gate, and PR-merge detection
- Added `ReviewState.DISMISSED` so a dismissed review no longer reads as an active review requesting changes, and closed a merge-vs-pending-CI race in `human_review_gate`
- `pr_state.py`'s `pull_requests` map is re-keyed per PR (`repo:number`, with a URL-keyed fallback for a PR recorded before its number was known) instead of per-repo, so concurrent PRs on the same repo no longer collide on one slot; `ci_evaluator` now shares the same lookup helper instead of duplicating the key logic


src/forge/integrations/source_control/github/adapter.py
```diff
@@ -0,0 +1,956 @@
...
+                if conclusion is not CheckConclusion.NONE
+                else CheckStatus.IN_PROGRESS
+            )
+
+        # Only a GitHub Actions-created check run's id doubles as an Actions
+        # job id. Other Apps (CodeQL, Codecov, ...) also set a numeric id,
+        # so id-presence alone isn't enough -- gate on the app slug too.
+        entry_id = entry.get("id")
+        is_actions_check = (entry.get("app") or {}).get("slug") == _GITHUB_ACTIONS_APP_SLUG
+        logs_url = str(entry_id) if entry_id is not None and is_actions_check else None
```
eshulman2: **P1 — a check-run ID is not an Actions job ID**

This stores the Checks API resource ID as though it were an Actions workflow job ID. The two IDs address different resources: check-run IDs are used with `/check-runs/{check_run_id}`, while `get_check_logs` calls `/actions/jobs/{job_id}/logs`. The `github-actions` app slug identifies the producer but does not make those IDs interchangeable, so real log retrieval will generally return 404.

Please resolve the corresponding Actions job and store/use its actual job ID.

Could you also add coverage that:

- uses deliberately different check-run and workflow-job IDs and asserts the job ID reaches `get_job_logs`;
- covers no matching job;
- covers ambiguous or multiple matching jobs;
- uses realistic GitHub check-run, workflow-run, and workflow-job response fixtures so mocks cannot accidentally encode the same invalid ID assumption?

References: [Checks API](https://docs.github.com/en/rest/checks/runs) and [Actions job logs API](https://docs.github.com/en/rest/actions/workflow-jobs).

danchild: addressed

eshulman2:
Review follow-up: `GitHubClient.get_reviews()` requests `per_page=100` but does not follow pagination, so `GitHubAdapter.get_review_threads()` silently omits review submissions after the first page. On long-lived PRs with more than 100 submissions, this can hide newer approval/change-request/dismissal decisions. Could this follow GitHub’s `Link` pagination until exhausted and include a regression test with a second page?

eshulman2:
### Code review

Found 1 issue:

1. `GitHubClient.get_reviews()` requests only `per_page=100` and does not follow GitHub's Link-header pagination, so on a PR with more than 100 review submissions the later reviews are silently dropped. `GitHubAdapter.get_review_threads()` builds approval / changes-requested / dismissal state from this list, so review decisions past the first page would be missed. This also confirms the earlier reviewer comment on this PR, which asked for Link pagination plus a regression test; the same unpaginated pattern appears in `get_review_comments()` (L462) and the review-threads REST fallback (L437).

https://github.com/forge-sdlc/forge/blob/0f9e1f2f385cfa6bb6a27c57f289ca93b7ebe694/src/forge/integrations/github/client.py#L188-L210

